PR #9, opened 2019-09-11
[README] worker:prod rebuild is not triggered automatically on packit:prod rebuild

jpopelka:
:-(

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,10 +67,11 @@ There are separate images for the [service / web server](https://hub.docker.com/
 There are separate images for staging and production deployment.
 Staging images are `:stg` tagged and built from `master` of `packit` and `packit-service`.
 Production images are `:prod` tagged and built from `stable` branch of `packit` and `packit-service`.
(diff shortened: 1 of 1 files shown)


TomasTomecek:
I thought that you can set this in docker hub - rebuild if base changes

jpopelka:
Yes, that's what I had set, but yesterday I was investigating why there are no new `packit-service-worker:prod` builds even when we moved `packit`'s `stable` branch and realized that the info next to "Enable for Base Image" [here](https://cloud.docker.com/u/usercont/repository/docker/usercont/packit-service-worker/builds/edit) says:
"Base images are specified in the `FROM:` directive in the Dockerfile of the source directory."

I'm not sure how to interpret that.
We have [Dockerfile](https://github.com/packit-service/packit-service/blob/master/Dockerfile), [Dockerfile.worker](https://github.com/packit-service/packit-service/blob/master/Dockerfile.worker) and [Dockerfile.worker.prod](https://github.com/packit-service/packit-service/blob/master/Dockerfile.worker.prod) and if the help says that only the first one is taken into account, then it means that auto-rebuild-when-base-changes does not work for worker:[prod|stg] :-(

PR #10, opened 2019-09-12
current prod configuration

TomasTomecek:
* increase memory for redis
* hello PR: add user.login
* we don't have quota for 3 replicas
* worker, readiness check: use status


TomasTomecek:
the status thing was in the previous PR I'd say and I didn't rebase properly or aliens

https://github.com/packit-service/deployment/blob/master/openshift/deployment-worker.yml.j2#L97

PR #21, opened 2019-10-07
deploy: make it work with ansible 2.7 & 2.8

TomasTomecek:
didn't try it though, should just work™

### DIFF
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -36,12 +36,6 @@
     image_worker: ""
     image_worker_digest: ""
   tasks:
-    - name: require ansible >= 2.8
-      assert:
-        that:
-        - 'ansible_version.minor >= 8'
-        fail_msg: 'Please upgrade to ansible 2.8 or later.'
-
     - name: include variables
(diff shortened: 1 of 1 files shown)


rpitonak:
> Let's test in production then

[Tested](https://softwarefactory-project.io/zuul/t/local/build/5351de638f304287ae2db85a59be5b88) and unfortunately does not work.

jpopelka:
> Tested and unfortunately does not work.

I've just noticed that [WE install ansible-2.7.10-1.fc30](https://github.com/packit-service/packit-service-zuul/blob/master/files/deploy-service.yaml#L10) - isn't that the cause why Zuul uses ansible-2.7? Wouldn't changing that to just `ansible` (unversioned, i.e. latest, i.e. 2.8.5) fix it too?

rpitonak:
@jpopelka yes, you identify the problem but it is actually [here](https://github.com/packit-service/packit-service/blob/72b842218c1314ae2d90f0feb8a49ebdcc7bdafe/files/deployment.yaml#L10)

jpopelka:
@rpitonak should I open a PR with the above-suggested change then?

PR #37, opened 2019-12-12
add postgresql

TomasTomecek:
tested locally and it werks

### DIFF
--- a/openshift/pg.yml
+++ b/openshift/pg.yml
@@ -1,0 +1,117 @@
+# MIT License
+#
+# Copyright (c) 2018-2019 Red Hat, Inc.
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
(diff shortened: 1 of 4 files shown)


openshift/pg.yml
```diff
@@ -0,0 +1,117 @@
...
+        - name: POSTGRESQL_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              key: database-password
+              name: postgresql
+        - name: POSTGRESQL_DATABASE
+          valueFrom:
+            secretKeyRef:
+              key: database-name
+              name: postgresql
```
jpopelka:
```yaml
key: postgres-secret
name: database-name
```

playbooks/deploy.yml
```diff
@@ -93,6 +96,7 @@
       loop:
         - ../openshift/redis.yml
         - ../openshift/service.yml
+        - ../openshift/pg.yml
```
jpopelka: aha, you've saved some letters, nice work  ;-)

openshift/pg.yml
TomasTomecek: yup, just ran into this while playing with it locally

playbooks/deploy.yml
TomasTomecek: yah, I'm lazy :D

openshift/pg.yml
jpopelka: yeah, and I switched the keys  :man_facepalming:

PR #42, opened 2020-01-10
utilize image and image_worker vars in object defs

TomasTomecek:


openshift/statefulset-worker.yml.j2
```diff
@@ -67,7 +67,7 @@ spec:
           secretName: packit-config
       containers:
       - name: packit-worker-1
-        image: packit-worker:{{ deployment }}
+        image: "{{ image_worker }}:{{ deployment }}"
```
jpopelka: Nope, you got it wrong.

We want this to be hardcoded to the [name:tag of the image stream](https://github.com/packit-service/deployment/blob/master/openshift/imagestream-worker.yml.j2#L27).
[There is the place](https://github.com/packit-service/deployment/blob/master/openshift/imagestream-worker.yml.j2#L33) to use the `image_worker` var, so that the image stream will import your image(s) if you re-define the var.

jpopelka:
Feel free to reopen and add some comments into the `yaml`s.

openshift/statefulset-worker.yml.j2
TomasTomecek: I don't get it >_>

TomasTomecek: anyone, in order to make this work locally, I had exterminate the triggers

jpopelka: ok, then we probably need to discuss this in person

PR #43, opened 2020-01-10
HOWTO Prod re-deploy

jpopelka:


jpopelka:
> does `make deploy` imply `make import-images`?

I'm not sure. I think when I was working on the migration to image streams it was working locally, so I guess the image stream does some initial import.
But you're right, in case of Openshift Online we should probably run it anyway.

#44

PR #45, opened 2020-01-28
[Image streams] set 'importPolicy:scheduled' conditionally

jpopelka:
to `false` on `prod` and `true` otherwise.

This doesn't change anything at the moment since these imports don't work on Online anyway.
But will be needed once we move elsewhere.


TomasTomecek:
could you please document what the option means?

jpopelka:
PTAL

TomasTomecek:
Oh, now I get it - my brain doesn't work much lately.

PR #49, opened 2020-02-03
[README] local development

jpopelka:


README.md
```diff
@@ -117,6 +108,50 @@ And see the `packit-worker-x` pods being re-deployed from the older image.
 3. Import images -> re-deploy
   - If you don't want to wait for [it to be done automatically](#continuous-deployment) you can [do that manually](#manually-import-a-newer-image)
 
+### Local development A.K.A. how do I test my changes?
```
csomh: Maybe simplify this title. Helps readers to be able to scan the page faster.

```suggestion
### How do I test my changes?
```

README.md
```diff
@@ -117,6 +108,50 @@ And see the `packit-worker-x` pods being re-deployed from the older image.
...
+
+Below are 3 ways how to test the service.
```
csomh: If you remove this line now, we won't need to update the number in the future :nerd_face:

README.md
```diff
@@ -117,6 +108,50 @@ And see the `packit-worker-x` pods being re-deployed from the older image.
...
+
+#### docker-compose (quick & dirty)
+
+There's a [docker-compose.yml in packit-service](https://github.com/packit-service/packit-service/blob/master/docker-compose.yml).
+You have to prepare `secrets/dev/` there to make it work, but then just `docker-compose up` should start whole service and the API should be accessible at https://localhost:8443/api/
+
+Things get more tricky if you want to
```
csomh: Nitpick: not sure about what are the habits for formatting markdown files, I usually find it good to make them readable even when they are viewed as source.

README.md
```diff
@@ -117,6 +108,50 @@ And see the `packit-worker-x` pods being re-deployed from the older image.
...
+  - put http://<the-public-url>/webhooks/github into Webhook URL
+- [install](https://developer.github.com/apps/installing-github-apps/) it @ https://github.com/settings/apps/<your-github-app>/installations to some repository
+- do some action (issue, PR) in that repository and see [ngrok web ui](http://localhost:4040)
+
+#### oc cluster up (slow & better)
+
+Because we run the service in OpenShift the more reliable way to test it is to run an Openshift cluster locally and deploy the service there.
+`oc cluster up` spawns the Openshift cluster.
+Create `secrets/dev/` (steal them from our secret repo).
+`cd vars; cp template.yml dev.yml` and in `dev.yml` set `host: https://127.0.0.1:8443` and `api_key:`.
```
csomh:
```suggestion
`cd vars; cp template.yml dev.yml` and in `dev.yml` set `host: https://127.0.0.1:8443` and `api_key` to the output of `oc whoami -t`.
```

README.md
```diff
@@ -117,6 +108,50 @@ And see the `packit-worker-x` pods being re-deployed from the older image.
...
+
+Run `DEPLOYMENT=dev make deploy`.
+
+#### Staging (quick & reliable & we'll kill you if you break it)
```
csomh: I find the wording in this section a little bit threatening :smile: Maybe update it to make it more friendly, but make the reader aware, that this is a shared environment, so it should be treated with care.

Also: what about the hourly updates? Wouldn't they override the changes done manually?

README.md
jpopelka: Which formatting issues are you referring to? (Not) breaking long lines?

csomh: Yes, not breaking long lines or breaking long lines at unexpected places. For example this paragraph would look like bellow when using `fmt`:

```text
Things get more tricky if you want to test that it works with a Github
App, because for that the service API needs to be publicly accessible
(so the App can send webhooks to it).
```

jpopelka: Ok, I see. I'm currently looking at adding some more linting/formatting tools into [pre-commit config](https://github.com/packit-service/deployment/blob/master/.pre-commit-config.yaml). Let me merge this as is and I'll fix it either with some formatter or manually in other PR.

README.md
jpopelka: > what about the hourly updates? Wouldn't they override the changes done manually?

only if there was a packit-service PR merged meanwhile

PR #56, opened 2020-02-13
fix pg deployment:

TomasTomecek:
* fix selectors
* add env vars how to conn to pg for service and worker


TomasTomecek:
rebased & merging

Issue #61, opened 2020-02-18
Enable individual deployment of all the services


csomh:
#60 Started with breaking out the deployment of Redis and Redis Commander. Would be nice to have the same mechanism for all the other services.

PR #62, opened 2020-02-19
templates reworked, so one template can contain multiple yaml sections

sakalosj:
partially implements #61

@all
if possible perform local test deployment

@csomh i had to comment out section  `name: Create redis-commander secrets` in deploy.yml, it was failing because of undefined variable

> ~~MSG:~~
> 
> ~~'redis_commander_user' is undefined~~

~~will fix it with your help.~~
redis_commander fixed


csomh:
Why was the ".j2" suffix removed from template files? I found it good to have it, b/c it was easier to tell which file is a template and which one a complete OpenShift resource definition. I also *think* using the ".j2" suffix is an [Ansible best practice](https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html), but I might be wrong.

openshift/cmap-packit-httpd-conf.yml
```diff
@@ -0,0 +1,55 @@
+# MIT License
```
csomh: Please make sure that all files which were renamed were renamed with `git mv`. The diff becomes very difficult to understand otherwise.

openshift/packit-service-fedmsg.yml
```diff
@@ -90,3 +90,20 @@ spec:
...
   strategy:
     type: Recreate
+---
+kind: ImageStream
+apiVersion: v1
+metadata:
+  name: packit-service-fedmsg
+spec:
+  tags:
+    - name: { { deployment } }
```
csomh: Let's stay consistent in using `{{ word }}`, and remove the space from between the curly brackets.

openshift/cmap-packit-httpd-conf.yml
sakalosj: i was not renaming files i was deleting and creating new one, didn't expect that git will create relation.
will be more careful next time

openshift/packit-service-fedmsg.yml
sakalosj: its done by pre-commit hook, dont know why. Its probably because in ansible yaml it means jinja2 variable and in "normal" yaml file doesnt have this, therfore it thinks it is dict in dict

sakalosj: will fix it by enclose it in double quotes

sakalosj:
> Why was the ".j2" suffix removed from template files? I found it good to have it, b/c it was easier to tell which file is a template and which one a complete OpenShift resource definition. I also _think_ using the ".j2" suffix is an [Ansible best practice](https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html), but I might be wrong.

you are right will change it

playbooks/prepare-local.yml
```diff
@@ -0,0 +1,50 @@
+- name: some config for local openshift deployment(oc have to be installed)
+  hosts: all
+  vars:
+    deployment: "{{ lookup('env', 'DEPLOYMENT') }}"
+  tasks:
+    - name: testing deployment var
```
csomh: Why is this useful?

Maybe it would make sense to default to `dev` 3 lines above?

playbooks/prepare-local.yml
```diff
@@ -0,0 +1,50 @@
...
+      lineinfile:
+        path: ../vars/{{ deployment }}.yml
+        regexp: "^project"
+        line: "project: packit-{{ deployment }}"
+
+    - name: update host
+      lineinfile:
+        path: ../vars/{{ deployment }}.yml
+        regexp: "^host"
+        line: "host: https://127.0.0.1:8443"
```
csomh: This is not always true, when using minishift this is going to be different. Maybe grep the host name from `oc status`?

sakalosj:
> `playbooks/prepare-local.yml` doesn't really has anything to do with reworking the other files.
> 
> Maybe it would be good to move it to a separate commit, and maybe even to open a separate PR for it, to have better visibility. The README should be also updated if we add this.

no it is working version of ansible script i am using to prepare env variables - will remove it from  PR

playbooks/prepare-local.yml
sakalosj: currently is true, can be improved in future

playbooks/prepare-local.yml
sakalosj: you want to test different deployments

sakalosj: playbooks/prepare-local.yml removed from PR

playbooks/prepare-local.yml
sakalosj: playbooks/prepare-local.yml removed from PR

playbooks/prepare-local.yml
csomh: For consistency, also from the `Makefile`, please :)

csomh:
@sakalosj is this still WIP?

sakalosj:
> @sakalosj is this still WIP?

no, i thought i changed it

playbooks/prepare-local.yml
sakalosj: ok done ;)

playbooks/deploy.yml
```diff
@@ -70,51 +70,58 @@
...
         host: "{{ host }}"
         api_key: "{{ api_key }}"
         verify_ssl: "{{ validate_certs }}" # 2.7 compatibility
-      loop:
-        - "{{ lookup('template', '../openshift/deployment.yml.j2') | from_yaml }}"
-        - "{{ lookup('template', '../openshift/statefulset-worker.yml.j2') | from_yaml }}"
+      tags:
+        - packit-service
+
+    - name: Deploy packit-service-worker (need to be process)
```
jpopelka: I think we don't need to use these `(need to be process)` vs. `(no need to process)` because we don't use the `from_yaml` anywhere, do we?

TomasTomecek:
what's the status here?

sakalosj:
is done but waiting for @csomh approval as he had most objections

sakalosj:
just noticed that there is also comment from @jpopelka, will remove unnecessary comments

playbooks/deploy.yml
sakalosj: i misunderstood meaning of that comment, removed

sakalosj:
@csomh please check and if you will find anything will handle it in new pull request

PR #63, opened 2020-02-26
don't do r-c secrets when not deploying it

TomasTomecek:


TomasTomecek:
merging, I need this ^

PR #66, opened 2020-03-10
Multiple changes:

TomasTomecek:
should I split to multiple PRs?

changes:
* deploy: wait for worker and service to be up and running
* have a default value for without_dashboard
* we have moved the httpd config to p-s


openshift/packit-service.yml.j2
```diff
@@ -76,9 +73,6 @@ spec:
               mountPath: /secrets
             - name: packit-config
               mountPath: /home/packit/.config
-            - name: httpd-conf
-              mountPath: /etc/httpd/conf.d/httpd-packit.conf
```
csomh: I asked this on packit-service/packit-service#473, but maybe went unnoticed: removing this here means that any update to `httpd-packit.conf` will require rebuilding the service image. Is this something we want?

(Configuration is kept in configmaps for a reason: it allows quick tweaks, as opposed to having to wait for an image build to finish.)

playbooks/deploy.yml
```diff
@@ -29,6 +29,7 @@
     without_fedmsg: false
     without_redis_commander: false
     without_flower: false
+    without_dashboard: true  # deploy w/o dashboard by default
```
csomh: Would be nice to update this in `vars/template.yml` too, so that this is the default when folks create their `vars/dev.yml`

openshift/packit-service.yml.j2
TomasTomecek: > I asked this on [packit-service/packit-service#473](https://github.com/packit-service/packit-service/pull/473), but maybe went unnoticed: removing this here means that any update to `httpd-packit.conf` will require rebuilding the service image. Is this something we want?

we almost don't update the config, so I don't see this as a problem - we rebuild the image regularly

> (Configuration is kept in configmaps for a reason: it allows quick tweaks, as opposed to having to wait for an image build to finish.)

^

The main point of this change is that having the config in a different repo makes local development of p-s very very hard. Once we bake it in, people can easily develop p-s. And as I said, I don't think we have changed the config in past ~4 months. Actually, it should be now even more trivial to change it, since I can edit the config, rebuild image and run it locally w/o the need of having openshift. Good job me! :D

playbooks/deploy.yml
TomasTomecek: oh, good point!

PR #71, opened 2020-03-16
added centos_msg pod config

sakalosj:
tested - pod is running in stg environment

### DIFF
--- a/openshift/packit-service-centosmsg.yml.j2
+++ b/openshift/packit-service-centosmsg.yml.j2
@@ -1,0 +1,95 @@
+# MIT License
+#
+# Copyright (c) 2018-2019 Red Hat, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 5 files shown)


openshift/packit-service-centosmsg.yml.j2
```diff
@@ -0,0 +1,95 @@
...
+  name: packit-service-centosmsg
+spec:
+  template:
+    metadata:
+      labels:
+        name: packit-service-centosmsg
+        app: packit
+    spec:
+      volumes:
+        - name: packit-secrets
```
csomh: While this will work, I'm a little bit worried of storing unrelated stuff in the same secret.

For one, this way the `packit-service-centosmsg` pod will be coupled to the `packit-service` one through this shared secret.

Two: this opens the door for `packit-service` to have access to centos messaging secrets, which is absolutely not required.

vars/template.yml
```diff
@@ -47,6 +47,9 @@ without_fedmsg: true
 # don't deploy redis-commander by default
 without_redis_commander: true
 
+# don't deploy centosmsg dc; good for local testing
+without_centosmsg: true
```
csomh: I think this should be also defined and set to some default value in `playbooks/deploy.yml#L32`

openshift/packit-service-centosmsg.yml.j2
sakalosj: ok will create issue for this, not right time for this now

csomh: Why not? You could create a new secret for the centos messaging secrets now and use that. You definitely shouldn't start to split up everything right now.

vars/template.yml
sakalosj: fixed

sakalosj: fixed

openshift/packit-service-centosmsg.yml.j2
sakalosj: thought it will require more time - fixed

sakalosj:
> LGTM, which repo is the `listen-to-centos-messaging` code in?

@jpopelka 
https://github.com/packit-service/packit-service-centosmsg

PR #77, opened 2020-04-14
Optionally Disable Sentry

IceWreck:
Will be valid only after packit-service/packit-service#561 is approved and merged.

- Adding `disable_sentry: True` to _secrets/packit-service.yaml_ disables sentry. 
- If this line is not present or `disable_sentry: False` is set then events will be sent to sentry like normal.

### DIFF
--- a/secrets/template/packit-service.yaml
+++ b/secrets/template/packit-service.yaml
@@ -20,7 +20,7 @@ validate_webhooks: False
 webhook_secret: <webhook_secret>
 testing_farm_secret: <testing_farm>
 
-disable_sentry: False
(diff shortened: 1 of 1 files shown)


secrets/template/packit-service.yaml
```diff
@@ -20,6 +20,8 @@ validate_webhooks: False
 webhook_secret: <webhook_secret>
 testing_farm_secret: <testing_farm>
 
+disable_sentry: False
```
TomasTomecek, line 23:
```suggestion
disable_sentry: True
```

PR #78, opened 2020-04-20
polish docs and code for get certs & dashboard

TomasTomecek:
* readme update
* get-certs now accepts domain instead of deployment

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,9 +242,8 @@ For more information please refer to [official docs](https://ansible.softwarefac
 
 ## Obtaining a Let's Encrypt cert using `certbot`
 
-Please beer (Milk coffee stout) in mind this is the easiest process I was able
-to figure out: there is a ton of places for improvements and ideally make it
-automated 100%.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -310,6 +309,21 @@ We'll use another session for that. First, copy the script into the pod:
...
+
+```
+Create a file containing just this data:
+
+<abc>
+
+And make it available on your web server at this URL:
+
+http://dashboard.stg.packit.dev/.well-known/acme-challenge/<xyz>
+```
```
jpopelka, line 324: - It says `dashboard.stg.packit.dev` while the rest of the text says `prod.packit.dev`.
- This is from the `certbot` output, isn't it? I'd suggest moving it up to where we say: "... until it tells you to create a file containing data abc and make it available on your web server at URL ..." (if you still insist on adding it - since it's a bit redundant with that sentence)

TomasTomecek: good point, will update

TomasTomecek: I added it b/c I was confused what abc and xyz are in the command below.

PR #83, opened 2020-04-27
2.9 is in all Fedoras/EPELs

jpopelka:


jpopelka:
> shouldn't we also change something in `vars/`?

nope, [vars/template.yml](https://github.com/packit-service/deployment/blob/master/vars/template.yml#L40) already uses `validate_certs`

jpopelka:
Why the heck is Zuul ignoring my PR?

jpopelka:
recheck

PR #84, opened 2020-04-29
[validation] Add a README, fix path to the copr config in cron job

lbarcziova:


haxxx/validation/openshift/job-run-validation.yaml
```diff
@@ -53,7 +53,7 @@ spec:
               command: ["python3", "/usr/bin/packit-service-validation.py"]
               volumeMounts:
                 - name: copr-config
-                  mountPath: ~/.config/
+                  mountPath: /.config
```
jpopelka, line 56: not sure about this, but I believe you :-)

lbarcziova: I am not sure about this either, but this I found in pod logs:
` There is no config file: /.config/copr`

PR #89, opened 2020-06-01
Set resources:limits:memory: to 80Mi and remove resources:requests:

jpopelka:
Services which need just few dozens Mi of mem were previously set to limit 128Mi.

After setting (#88) one of them to requests:32Mi & limits:64Mi, we were getting:
`spec.containers[0].resources.requests: Invalid value: '80Mi': must be less than or equal to memory limit`

When I'm trying to set requests to 64Mi or remove it completely the problem is still there,
so I think Openshift Online is set so that requests (& limits) can't be less than 80Mi.

EDIT: I'll test this on `stg` once I resolve a bug in `ansible-playbook` which prevents me from running `make deploy`.

### DIFF
--- a/haxxx/job-import-images.yml
+++ b/haxxx/job-import-images.yml
@@ -32,10 +32,7 @@ spec:
                       # name: importimager-token-gl9g9
                       key: token
               resources:
-                requests:
-                  memory: "64Mi"
-                  cpu: "100m"
                 limits:
-                  memory: "128Mi"
+                  memory: "80Mi"
                   cpu: "100m"
           restartPolicy: OnFailure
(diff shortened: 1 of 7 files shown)


jpopelka:
> you can easily test by editing the objects in webui

which is what I did (but only with the failing `centosmsg` dc) :-)
thanks

jpopelka:
I spent yesterday trying to make the ansible work - with no luck - it always stucks on `k8s` module (reported upstream).

@TomasTomecek since you're the 'deploy-master' this week :-) can I ask you to pull this change (I'm merging this now) and re-deploy stg? Just ping me before you're doing it so that I can see how it goes. Thanks.

PR #90, opened 2020-06-05
[playbooks/deploy.yml] push dev images with podman

jpopelka:
If you want to stay with docker, just put the following into `vars/dev.yml`:
```yaml
container_engine: "docker"
```

https://github.com/packit-service/packit-service/pull/669

Why?
I was wondering why when I run 'make check-inside-openshift' @ packit-service, my freshly added tests are not run - and, it's because the images are built with `podman` by default and pushed (some old images) with `docker`.

### DIFF
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -41,6 +41,9 @@
     path_to_secrets: "../secrets"
     # to be used in Image streams as importPolicy:scheduled value
     auto_import_images: "{{(deployment != 'prod')}}"
+    # used in dev/zuul deployment to tag & push images to cluster
+    container_engine: "podman"
+    tls_verify_false: "{{ '--tls-verify=false' if container_engine == 'podman' else '' }}"
   tasks:
     - name: include variables
       include_vars: ../vars/{{ deployment }}.yml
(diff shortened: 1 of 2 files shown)


jpopelka:
Strangely when I use it, the deployed worker's liveness probe fails - I was checking whether it's https://github.com/packit-service/packit-service/pull/663 and nope, the celery is 4.4.2 from PyPI and when I deploy without probes and try those celery status/inspect commands they pass with OK, so I'm clueless.

TomasTomecek:
> worker's liveness probe fails

logs please

TomasTomecek:
seems like it could be a race condition

jpopelka:
So eventually it all settles down and the worker is running, but the probe really failed at least once:
```
Type     Reason     Age   From                Message
  ----     ------     ----  ----                -------
Warning  Unhealthy  10s   kubelet, localhost  Readiness probe failed: 12:19:28.628 utils.py          
Error: No nodes replied within time constraint.

jpopelka:
So I've just seen the same (liveness probe failed) when using docker in p-s' `make check-inside-openshift` so it's most likely not related to this PR.

Can someone give thumbs up to this and https://github.com/packit-service/packit-service/pull/669 so that we can try in Zuul?

jpopelka:
Just FYI: 
I'm definitely seeing this with docker (without this PR) too. I've played with it a bit, tried few older celery versions, but still the same, i.e. the worker pod's readiness/liveness checks fail several times and after the pod is re-created couple times, something in celery starts to work and the `celery status/inspect` (and the health checks) finally succeed and the `Wait for worker-0 to be running` task & deployment finish.

PR #92, opened 2020-06-18
[playbooks/deploy.yml] set container_engine conditionally

jpopelka:
To be in sync with
https://github.com/packit-service/packit-service/blob/master/Makefile#L6


playbooks/deploy.yml
```diff
@@ -71,6 +70,7 @@
 
     - name: Push dev images to local registry
       block:
+        - set_fact: tls_verify_false="{{ '--tls-verify=false' if 'podman' in container_engine else '' }}"
```
TomasTomecek, line 73: so docker doesn't need the tls-verify thing?

jpopelka: nope

PR #95, opened 2020-06-22
New delete-PVCs CronJob to delete old PVCs in sandbox

jpopelka:
Work-around for https://github.com/packit-service/packit-service/issues/409


jpopelka:
Tested @ `stg`, ready for review

cron-jobs/delete-pvcs/delete-pvcs.sh
```diff
@@ -0,0 +1,16 @@
...
+set -x
+
+oc get pvc -n "packit-${DEPLOYMENT}-sandbox" | grep pvc |
+while IFS= read -r line; do
+  name=$(echo $line | awk '{print $1}')
+  age=$(echo $line | awk '{print $7}')
+  if [ "${age: -1}" == "d" ]; # days
+  then
+    echo "deleting ${name}"
+    oc delete pvc "${name}"
```
csomh, line 14: Please make the namespace explicit here too.

cron-jobs/delete-pvcs/delete-pvcs.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/bash
+
+oc login "${HOST}" --token="${TOKEN}" --insecure-skip-tls-verify
```
csomh, line 3: Why is `--insecure-skip-tls-verify` needed here?

csomh:
> what I remember when this is happening, we tend to run out of quota within hours, not days

Pretty much all the times I was cleaning up storage I was deleting PVCs which were days old. So I would say the current solution is good enough, but I'm not against having a more aggressive approach.

cron-jobs/delete-pvcs/delete-pvcs.sh
jpopelka: added

cron-jobs/delete-pvcs/delete-pvcs.sh
jpopelka: removed

TomasTomecek:
> > what I remember when this is happening, we tend to run out of quota within hours, not days
> 
> Pretty much all the times I was cleaning up storage I was deleting PVCs which were days old. So I would say the current solution is good enough, but I'm not against having a more aggressive approach.

oki, maybe things changed over time and the situation is better now (with days, not hours)

PR #106, opened 2020-07-15
Deploy script setupcfg2rpm.py

sakalosj:
Script setupcfg2rpm.py parses setup.cfg, extracts python packages from options->install_requires
section and wraps them into python3dist(package) so output can by used with dnf package installer.

    Example:
    $ dnf install $(setupcfg2rpm.py path_to_setupcfg)
    python3dist(some_package)
    python3dist(other_package)
    $

### DIFF
--- a/scripts/setupcfg2rpm.py
+++ b/scripts/setupcfg2rpm.py
@@ -1,0 +1,64 @@
+#!/usr/bin/python3
+# MIT License
+#
+# Copyright (c) 2019 Packit
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""
+Script setupcfg2rpm.py parses setup.cfg, extracts python packages from options->install_requires
+section and wraps them into python3dist(package) so output can by used with dnf package installer.
+
+    Example:
+    $ dnf install $(setupcfg2rpm.py path_to_setupcfg)
+    python3dist(some_package)
+    python3dist(other_package)
+    $
+
+"""
+
+import os
+import re
+import pathlib
+import argparse
+import configparser
+
+
+def normalize_name(name: str) -> str:
+    """https://www.python.org/dev/peps/pep-0503/#normalized-names"""
+    return re.sub(r'[-_.]+', '-', name).lower()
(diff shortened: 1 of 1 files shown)


scripts/setupcfg2rpm.py
```diff
@@ -0,0 +1,73 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description='Parse rpm requirements from setup.cfg from section '
+                    '[options], key install_requires')
+    parser.add_argument('path', type=str, help='Path to setup.cfg')
+
+    args = parser.parse_args()
+
+    setupcfg_path = pathlib.Path(f"{args.path}/setup.cfg")
```
csomh, line 61: Could the whole path be provided and skip this concatenation here?

Doing this here is not in line with the help text of the argument which says that `path` is "Path to setup.cfg".

scripts/setupcfg2rpm.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def normalize_name(name: str) -> str:
+    """https://www.python.org/dev/peps/pep-0503/#normalized-names"""
+    return re.sub(r'[-_.]+', '-', name).lower()
+
+
+def pydist_wrapp(name: str) -> str:
+    """Wrapps string name: name -> python3dist(name)"""
+    return f"python3dist({name})"
```
csomh, line 50: This seems to be a very short function. Could the f-string be used in-place instead?

scripts/setupcfg2rpm.py
```diff
@@ -0,0 +1,73 @@
...
+        description='Parse rpm requirements from setup.cfg from section '
+                    '[options], key install_requires')
+    parser.add_argument('path', type=str, help='Path to setup.cfg')
+
+    args = parser.parse_args()
+
+    setupcfg_path = pathlib.Path(f"{args.path}/setup.cfg")
+
+    if not setupcfg_path.is_file():
+        raise FileNotFoundError(f'"{setupcfg_path}" is not a file.')
```
csomh, line 64: This error with this condition could mislead the reader, as it would also raise a `FileNotFoundError` in case the path is directory.

As this is not a user facing script, I would just rely on whatever error `config.read` raises if the path is not OK.

scripts/setupcfg2rpm.py
```diff
@@ -0,0 +1,73 @@
...
+
+    config = configparser.ConfigParser()
+    config.read(setupcfg_path)
+
+    python_packages = config["options"]["install_requires"].splitlines()
```
csomh, line 69:
```suggestion
    python_packages = config["options"]["install_requires"].strip().splitlines()
```

And the you could remove the `if` below.

scripts/setupcfg2rpm.py
```diff
@@ -0,0 +1,73 @@
...
+
+    result = [pydist_wrapp(normalize_name(pkg_name)) for pkg_name in python_packages if pkg_name]
+
+    print(os.linesep.join(result))
```
csomh, line 73: Wouldn't it make more sense to separate these with space, as they are going to be used as arguments to `dnf install`?

sakalosj: it is normal for ux scripts to have line separated output - dnf consumes it without any issues and is easier to process by tools like grep, sed, etc. in case of need, so i will suggest to keep it as it is

csomh: Good points! Let's keep it this way then.

scripts/setupcfg2rpm.py
sakalosj: fixed

scripts/setupcfg2rpm.py
sakalosj: config read is accepting non existing files and directories without any error, so included test to check if provided path is directory

scripts/setupcfg2rpm.py
sakalosj: fixed

scripts/setupcfg2rpm.py
sakalosj: fixed

scripts/setupcfg2rpm.py
csomh: Okay, this might be too much of a nitpicking so please feel free to ignore. How about:

```python
with open("setup.cfg", "r") as fp:
     config.read_file(fp)
```

in which case `open` would raise meaningful errors.

sakalosj: good point - implemented

sakalosj:
> looks good

thanks, sorry for not waiting for your approval, thought you're busy and because @csomh gave me good feedback i decided to move forward instead of bothering you. same applies for @jpopelka and others

PR #107, opened 2020-07-16
Add configuration for running celery beat

lbarcziova:


openshift/packit-service-beat.yml.j2
```diff
@@ -0,0 +1,125 @@
...
+          name: packit-worker:{{ deployment }}
+  replicas: 1
+  strategy:
+    type: Recreate
+---
+# https://docs.openshift.com/container-platform/3.11/dev_guide/managing_images.html#writing-image-stream-definitions
+kind: ImageStream
+apiVersion: v1
+metadata:
+  name: packit-worker
```
jpopelka, line 113: This ImageStream is already created by the `packit-service-worker.yml.j2` and the beat will use it too.

openshift/packit-service-beat.yml.j2
```diff
@@ -0,0 +1,125 @@
...
+              mountPath: /home/packit/.config
+          command:
+            - "/usr/bin/run_worker.sh"
+          resources:
+            requests:
+              memory: "100Mi"
+              cpu: "50m"
+            limits:
+              memory: "200Mi"
+              cpu: "100m"
```
jpopelka, line 94: The beat doesn't need so much memory. If I understand it correctly it'll do nothing most of the time.
```suggestion
            limits:
              memory: "80Mi"
              cpu: "50m"
```

openshift/packit-service-beat.yml.j2
```diff
@@ -0,0 +1,125 @@
...
+              valueFrom:
+                secretKeyRef:
+                  key: database-user
+                  name: postgres-secret
+            - name: POSTGRESQL_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  key: database-password
+                  name: postgres-secret
+            - name: POSTGRESQL_DATABASE
```
jpopelka, line 72: I'm just wondering if we need to set all those variables from all the secrets.
The beat should IMHO not need any of those, but I'll leave this on you to try.

vars/stg_template.yml
```diff
@@ -46,6 +46,8 @@ api_key: ""
 
 # with_flower: true
 
+# with_beat: true
```
jpopelka, line 49: add it to `vars/template.yml` as well, please

openshift/packit-service-beat.yml.j2
```diff
@@ -0,0 +1,125 @@
+# MIT License
+#
+# Copyright (c) 2018-2019 Red Hat, Inc.
```
lachmanfrantisek, line 3:
```suggestion
# Copyright (c) 2020 Red Hat, Inc.
```

openshift/packit-service-beat.yml.j2
lbarcziova: When I was trying to run it with this configuration, beat had problems to start so I made the limits higher, is it okay?

jpopelka: ok

openshift/packit-service-beat.yml.j2
lbarcziova: As the packit_service.worker.tasks is called and it imports the app from cerelizer.py, I think the redis and postgres env vars are needed.

PR #116, opened 2020-08-10
Rename Copr project name in validation script

lbarcziova:
After the namespace renaming `packit-service` -> `packit`, the Copr project name for hello-world PRs is different as well, so the script was trying to find the wrong Copr projects.

### DIFF
--- a/cron-jobs/validation/packit-service-validation.py
+++ b/cron-jobs/validation/packit-service-validation.py
@@ -30,7 +30,7 @@ from ogr.services.github import GithubService
 
 copr = Client.create_from_config_file()
 service = GithubService(token=getenv("GITHUB_TOKEN"))
-project = service.get_project(repo="hello-world", namespace="packit-service")
+project = service.get_project(repo="hello-world", namespace="packit")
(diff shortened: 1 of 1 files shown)


cron-jobs/validation/packit-service-validation.py
```diff
@@ -89,7 +89,7 @@ def check_build_submitted(self):
         Check whether the build was submitted in Copr in time 30 minutes.
         :return:
         """
-        project_name = f"packit-service-hello-world-{self.pr_id}"
+        project_name = f"packit-hello-world-{self.pr_id}"
```
csomh, line 92: This seems to be the Copr project name. Does it need to change?

lbarcziova: Yes, since copr project name is based on namespace and repo name of the project, e.g. https://copr.fedorainfracloud.org/coprs/packit/packit-hello-world-108/ (old project - https://copr.fedorainfracloud.org/coprs/packit/packit-service-hello-world-108/)

PR #124, opened 2020-08-18
Add secrets/secrets.yaml to .gitignore

lbarcziova:


.gitignore
```diff
@@ -8,3 +8,5 @@ vars/*.yml
 secrets/prod
 secrets/stg
 secrets/dev
+
+secrets/secrets.yaml
```
jpopelka, line 12: What do you think about gitignoring whole secrets/?
```suggestion
secrets/*
```

lbarcziova: But we have there the templates and README.md

jpopelka: You're right. Go ahead & merge.

PR #125, opened 2020-08-19
Add env vars to be able to run Sentry in tokman, p-s

lbarcziova:
- add `SENTRY_SECRET`, `DEPLOYMENT` vars to tokman env
- add `SENTRY_SECRET` var to packit-service env (packit/packit-service#784 - I see `configure_sentry()` is called also in service, but the env var is not set)

### DIFF
--- a/openshift/packit-service.yml.j2
+++ b/openshift/packit-service.yml.j2
@@ -68,6 +68,11 @@ spec:
                 secretKeyRef:
                   key: database-name
                   name: postgres-secret
+            - name: SENTRY_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: packit-sentry
+                  key: sentry
           volumeMounts:
             - name: packit-secrets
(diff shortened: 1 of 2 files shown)


openshift/packit-service.yml.j2
```diff
@@ -68,6 +68,11 @@ spec:
                 secretKeyRef:
                   key: database-name
                   name: postgres-secret
+            - name: SENTRY_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: packit-sentry
+                  key: sentry
```
jpopelka, line 75: Thank you!

Do you think this will fix https://github.com/packit/packit-service/issues/784 ?

jpopelka (approved):

openshift/packit-service.yml.j2
lbarcziova: Yes, I think it should

PR #135, opened 2020-09-11
[worker] Run probes only if not using AWS SQS

jpopelka:
[SQS does not support broadcast](https://docs.celeryproject.org/en/latest/getting-started/brokers/index.html#broker-overview) and `celery status/inspect` results in `Error: Broadcast not supported by transport 'sqs'`

### DIFF
--- a/openshift/packit-service-worker.yml.j2
+++ b/openshift/packit-service-worker.yml.j2
@@ -139,7 +139,8 @@ spec:
               command:
                 - bash
                 - -c
-                - "celery -A $APP status | grep $(cat /etc/hostname) | grep OK"
+                # if we don't use AWS SQS (doesn't support this)
+                - 'if (test -z "$AWS_ACCESS_KEY_ID"); then celery -A $APP status | grep $(cat /etc/hostname) | grep OK; else true; fi'
             initialDelaySeconds: 10
             periodSeconds: 60
           livenessProbe:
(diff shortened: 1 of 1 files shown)


csomh:
Is there any way to get probes with similar effect with SQS?

Would it be possible to not comment out these probes before we have a final commitment to SQS?

jpopelka:
> Is there any way to get probes with similar effect with SQS?

I haven't found any. [SQS doesn't support broadcasting/remote inspection](https://docs.celeryproject.org/en/latest/getting-started/brokers/index.html#broker-overview).

But my feeling is that these probes (`celery status/inspect`) never worked reliably.
I don't think I filled any issue, but I remember investigating worker (re)starting because the probe was claiming it's not ready (yet) and needed several (re)starts to finally succeed. It was in requre-tests/Zuul and I don't know whether it's still happening, but I know that sometimes it takes a looong time before the [Wait for worker-0 to be running](https://github.com/packit/deployment/blob/master/playbooks/deploy.yml#L328) task passes when I run it locally.

Also @TomasTomecek might have a say since he implemented those initially and I think they were also commented out at some point in the history.

> Would it be possible to not comment out these probes before we have a final commitment to SQS?

sure

jpopelka:
> Would it be possible to not comment out these probes before we have a final commitment to SQS?

or I can run them conditionally only if we don't use SQS:
`if (test -z $AWS_ACCESS_KEY_ID); then celery -A $APP status | grep $(cat /etc/hostname) | grep OK; else true; fi`

csomh:
> > Would it be possible to not comment out these probes before we have a final commitment to SQS?
> 
> or I can run them conditionally only if we don't use SQS:
> `if (test -z $AWS_ACCESS_KEY_ID); then celery -A $APP status | grep $(cat /etc/hostname) | grep OK; else true; fi`

I like this approach better, so :+1:

csomh (approved):
Please update the PR title and description before merging (as it will be used for the merge commit).

PR #140, opened 2020-09-16
Document the pushgateway implementation

lbarcziova:


jpopelka (approved):
Thank you!

jpopelka:
The reason why I asked for this is that I need to understand what'd need to be changed if/when we move workers to PSI.
How does the pushgateway actually collect metrics from the workers?

lbarcziova:
> The reason why I asked for this is that I need to understand what'd need to be changed if/when we move workers to PSI.
How does the pushgateway actually collect metrics from the workers?

This is not implemented yet, but [here](https://github.com/prometheus/client_python#exporting-to-a-pushgateway) you can find how to export to the Pushgateway. The current implementation assumes that workers and pushgateway are in the same cluster, since only GET requests are allowed for public `workers.packit.dev/metrics`, but later it can be changed to use basic authentication and allow also POST requests for the route if needed.

PR #144, opened 2020-09-24
Use Prometheus Aggregation Gateway instead of Pushgateway

lbarcziova:
Since the [counter for Copr builds](https://github.com/packit/packit-service/pull/827) is rewritten with each push (so the value is always 1, see [metrics](https://workers.stg.packit.dev/metrics)) when using Pushgateway, I checked for similar issues and found [Prometheus Aggregation Gateway](https://github.com/weaveworks/prom-aggregation-gateway) which could better fit our use case. Tried it locally, it doesn't support the `persistence.file` and `persistence.interval` args, but otherwise worked well.

### DIFF
--- a/openshift/pushgateway.yml.j2
+++ b/openshift/pushgateway.yml.j2
@@ -15,24 +15,14 @@ spec:
         name: pushgateway
         app: packit
     spec:
-      volumes:
-        - name: pushgateway-data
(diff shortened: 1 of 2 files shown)


jpopelka (approved):

csomh (approved):
:partying_face:

PR #158, opened 2020-12-03
Add script for moving stable branches

mfocko:

Dependency: `click`

As a remote uses: `github.com:namespace/repo.git` (expects SSH key to be set up)

1. `scripts/move_stable.py init` creates a directory and clones all git repositories with stable branch into it (ran only once)
   - resolves ambiguity of remotes `upstream` vs `origin` vs _any-other-creative-name_
   - all repos have same prefix for their path
2. `scripts/move_stable.py move-all`
   - runs git fetch
   - prints out log from stable to master (there's also hash of stable in case of removing the confirmation before moving)
   - automatically takes hash from master, but allows to input hash manually (I should add checking for invalid hash too)

- can be run from either the root of `deployment` or `scripts` directory (not interchangeable, `init` is specific to the directory), directory with repositories added to gitignore
- there's also mock-up of `status` command, I've hit API rate limit without token instantly; used token, I wasn't patient enough to wait for results
   My intention was to take commit statuses of PRs with periodic checks (labelled them with `should pass`) and print / guess last commit hash that could've been tested and print all hung / failed commit statuses
- Also I plan to clean up the running of git

---
_almost moved the stable while testing_

### DIFF
--- a/scripts/move_stable.py
+++ b/scripts/move_stable.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python3
+# Copyright Contributors to the Packit project.
+# SPDX-License-Identifier: MIT
+
+import click
+
+# import ogr
+
+import os
+import subprocess
+from typing import List
+
+
+NAMESPACE: str = "packit"
+REPOSITORIES: List[str] = [
+    "packit",
+    "packit-service-fedmsg",
+    "packit-service-centosmsg",
+    "sandcastle",
+    "dashboard",
+    "tokman",
+    "packit-service",
+]
+# TESTING_REPOSITORY: str = "hello-world"
+STABLE_BRANCH: str = "stable"
+ROLLING_BRANCH: str = "master"
+
+
+@click.group()
+def cli() -> None:
+    pass
+
+
+@cli.command(short_help="Creates a directory with all the repositories")
+def init() -> None:
+    os.makedirs("repositories", exist_ok=True)
+
+    for repository in REPOSITORIES:
+        subprocess.run(
+            ["git", "clone", f"github.com:{NAMESPACE}/{repository}.git"],
+            cwd=f"{os.curdir}/repositories",
+        )
+
+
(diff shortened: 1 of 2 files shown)


scripts/move_stable.py
```diff
@@ -0,0 +1,143 @@
...
+    pass
+
+
+@cli.command(short_help="Creates a directory with all the repositories")
+def init() -> None:
+    os.makedirs("move_stable_repositories", exist_ok=True)
+
+    for repository in REPOSITORIES:
+        subprocess.run(
+            ["git", "clone", f"github.com:{NAMESPACE}/{repository}.git"],
```
jpopelka, line 40: I'm getting `Permission denied (publickey)` without the following:
```suggestion
            ["git", "clone", f"git@github.com:{NAMESPACE}/{repository}.git"],
```

scripts/move_stable.py
```diff
@@ -0,0 +1,143 @@
...
+            click.echo(
+                f"===> {ROLLING_BRANCH} and {STABLE_BRANCH} are even for {NAMESPACE}/{repository} => Skipping"
+            )
+            continue
+
+        get_git_log(path_to_repository, remote, stable_hash, main_hash)
+        click.echo()
+
+        new_stable_hash = click.prompt(
+            f"Enter new hash for {STABLE_BRANCH}", default=main_hash
```
jpopelka, line 68:
```suggestion
            f"Enter new hash for {STABLE_BRANCH}", default=main_hash[:7]
```

jpopelka:
I just used it :rocket:

scripts/move_stable.py
mfocko: interesting, must've been my `github.com` alias in SSH config, will fix

scripts/move_stable.py
mfocko: ux only ;) but will change

thanks for review :rocket:

jpopelka (approved):

PR #166, opened 2021-01-11
Create a separate Makefile for the validation

lbarcziova:


TomasTomecek (approved):
LGTM, even though it would be nicer if the license header change would be done in separate commits

jpopelka (approved):

PR #172, opened 2021-01-26
worker memory: set to 896Mi

TomasTomecek:
896 = 512 + 256 + 128

cloning repos is memory intensive: glibc needs 300M+, kernel 600M+
during cloning, we need to account for git and celery worker processes


lachmanfrantisek (approved):

TomasTomecek:
this is still not enough

I just patched stg worker to use 1.2G and we are trying that

PR #181, opened 2021-03-04
[.github/workflows/base-image-rebuild] run on push to main

jpopelka:
so that the base image is rebuilt also when its Dockerfile changes.

Use organization level secrets instead of local repo secrets.

### DIFF
--- a/.github/workflows/base-image-rebuild.yml
+++ b/.github/workflows/base-image-rebuild.yml
@@ -3,11 +3,16 @@ on:
   schedule:
     # At 00:00 on Monday.
     - cron: "0 0 * * 1"
+  push:
+    branches:
+      - main
+    paths:
+      - containers/Dockerfile.base
 
 jobs:
   build:
     # To not run in forks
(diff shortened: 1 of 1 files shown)


csomh (approved):
Looks good! One suggestion: you could rephrase the commit message subject line to something like:

```
Build the base image when the Dockerfile is changed
```

Although the current subject line contains the same information, still 
1) It does it in a somewhat cryptic way, aka: this is not how humans communicate.
2) Unnecessarily duplicates almost the entire path of the file changed (and so makes the subject like way too long); this information can be easily seen from a `git log`, no need to repeat it.
3) Uses square brackets, which requires and extra effort to read in an email context, as it blends with the square-bracketed prefix added by GitHub. This is what I see in Gmail:
image
but this feels much better:
image
Alternatively if you feel the need to use a suffix to indicated context, I would find something like: `github: Build the base image when the Dockerfile is changed` easier to read.

jpopelka:
ok

PR #195, opened 2021-04-07
delete-pvcs now deletes PVCs from any Completed pods

### LINKED ISSUE #192: More aggressive delete-pvcs cron job
Currently the cron job [runs every hour](https://github.com/packit/deployment/blob/main/cron-jobs/delete-pvcs/job-delete-pvcs.yaml#L7) and [deletes PVCs older than 1 day](https://github.com/packit/deployment/blob/main/cron-jobs/delete-pvcs/delete-pvcs.sh#L13).
Recently it happened that we hit a noncompute / storage quota and when checking PVCs in packit-prod-sandbox there were about dozen of PVCs no older than 2 hours.
So we should either delete PVCs older than 1-2hours or even better somehow check whether a PVC is used in any running pod and if not, delete it.

jpopelka:
Previously we deleted only PVCs older than a day.
Now we delete all PVCs associated with pods in the Completed state.

EDIT: And also delete pods older than a day.

Fixes #192

### DIFF
--- a/cron-jobs/delete-pvcs/README.md
+++ b/cron-jobs/delete-pvcs/README.md
@@ -1,6 +1,6 @@
-## Periodically deleting old persistentvolumeclaims (PVCs)
+## Periodically deleting old Persistent Volume Claims (PVCs)
 
-[It happens from time to time](https://github.com/packit-service/packit-service/issues/409) that leftover PVCs in a sandbox namespace consume all storage quota.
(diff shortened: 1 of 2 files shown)


TomasTomecek (approved):

jpopelka:
Merging.
I'll wait with re-deploying this so that it doesn't interfere with your work on packit/packit-service#1047 @TomasTomecek

PR #207, opened 2021-05-11
Update redis and postgres images

lbarcziova:


TomasTomecek:
Could we get a note to README about the database upgrade?

I can see sclorg [documented the upgrade process](https://github.com/sclorg/postgresql-container/tree/generated/12#upgrading-database-by-switching-to-newer-postgresql-image-version) as well though the blog post you linked is more thorough.

I wonder how this is gonna be executed, would be great to have a plan and discuss it :)

lbarcziova:
Changed to [the approach](https://github.com/sclorg/postgresql-container/tree/generated/12#upgrading-database-by-switching-to-newer-postgresql-image-version) pointed out by Tomas - image should support automatic upgrade when `$POSTGRESQL_UPGRADE` is defined. There are 2 options available for it - copy and hardlink. I set it to copy, since it is safer (but requires memory for the data directory copy).

openshift/postgres.yml
```diff
@@ -53,7 +34,9 @@ spec:
                 secretKeyRef:
                   key: database-name
                   name: postgres-secret
-          image: registry.access.redhat.com/rhscl/postgresql-10-rhel7
+            - name: POSTGRESQL_UPGRADE
+              value: copy
```
csomh, line 38: This should be set only once, right? Wouldn't it be better to populate this from Ansible. Then you could set it once in your vars yaml but keep it unset by default (for future deployments).

csomh: Though probably you could keep this change locally and never actually make it in the history of this repository (if it needs to be used only once).

PR #208, opened 2021-05-11
make: sugest to run 'import-images' with 'deploy'

jpopelka:
Quay.io seems to keep only the latest image build in each repo.
In case we build a new image and run 'make deploy' without
the image being imported first, we get 'Error: ImagePullBackOff'.
The workaround is to run 'make import-images' as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ AP := ansible-playbook -vv -c local -i localhost, -e ansible_python_interpreter=
 # https://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html#special-tags
 TAGS ?= all
 
+# Consider running 'make import-images' to avoid possible 'Error: ImagePullBackOff'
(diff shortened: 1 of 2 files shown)


TomasTomecek:
> Quay.io seems to keep only the latest image build in each repo.
> In case we build a new image and run 'make deploy' without the image being imported first, we get 'Error: ImagePullBackOff'.
> The workaround is to run 'make import-images' as well.

sounds like we wouldn't be able to roll back, right? would it make sense to tag 2 images: current and previous(-known-to-be-good)?

jpopelka:
> would it make sense to tag 2 images: current and previous(-known-to-be-good)?

Yeah, I was thinking how to do that but tagging the previous hadn't come to my mind, thanks.
Now just find an easy way how to add that into [our workflow](https://github.com/packit/.github/blob/main/workflow-templates/build-and-push.yml) :-)

PR #210, opened 2021-05-19
Add persistent volume for repository cache

lachmanfrantisek:
Add persistent volume to the worker for repository-cache.
    
* The only option on OpenShift Online is AWS EBS provisioning
   supporting only `ReadWriteOnce` access:
   https://docs.openshift.com/online/pro/architecture/additional_concepts/storage.html#pv-restrictions  
   Because of that, we need to have one repository-cache for each worker.
    
* The volume is mounted to /repository-cache path.
* For now, the content of the cache will be maintained (adding/updating) manually.

### DIFF
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -155,6 +155,18 @@
       tags:
         - packit-service
 
+    - name: Deploy repository cache PVCs for packit-workers that serves both queues
+      vars:
+        name: "packit-worker-{{ item }}"
+      k8s:
+        namespace: "{{ sandbox_namespace }}"
+        definition: "{{ lookup('template', '{{ project_dir }}/openshift/sandcastle-volumes-for-cache.yml.j2') }}"
+        host: "{{ host }}"
+        api_key: "{{ api_key }}"
+        validate_certs: "{{ validate_certs }}"
+      loop: "{{ range(0, workers_all_tasks)|list }}"
+      when: workers_all_tasks > 0
+
     - name: Deploy packit-worker to serve both queues
       vars:
         name: packit-worker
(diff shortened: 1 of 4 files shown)


jpopelka (commented):
ok

TomasTomecek (approved):
I wonder if we should only keep trying this on stg before we feel confident in bringing it to prod

also, do we have documentation for how to update the cache? (asking b/c `ReadWriteOnce` is scaring me)

lachmanfrantisek:
> I wonder if we should only keep trying this on stg before we feel confident in bringing it to prod

Empty cache on prod should not cause any problems. So we can try first on stage without any change in deployment.

> also, do we have documentation for how to update the cache? (asking b/c `ReadWriteOnce` is scaring me)

* If you mean the `Once` in the type, it means `The volume can be mounted as read-write by a single node.`
* And regarding the process, we can start with the manual approach as we've discussed.

lachmanfrantisek:
@jpopelka Can you please take a look? What we need is to define for each worker:

- one PVC in the prod/stage namespace and mount it
- one PVC in the sandbox namespace that will be mounted through sandcastle using `SANDCASTLE_REPOSITORY_CACHE_VOLUME` env var of the worker.

playbooks/deploy.yml
```diff
@@ -170,6 +170,18 @@
         - packit-worker
       when: workers_all_tasks > 0
 
+    - name: Deploy repository cache PVCs for packit-workers that serves both queues
```
jpopelka, line 173: I'd deploy it before the workers, just to be sure it's ready when needed.

jpopelka (approved):
LGTM with a suggestion

playbooks/deploy.yml
lachmanfrantisek: Fixed.

TomasTomecek (approved):

lachmanfrantisek:
I am dropping the `Configure repository-cache volume for sandcastle` commit, this needs to be done elsewhere: https://github.com/packit/packit-service/issues/1170

PR #212, opened 2021-05-24
[validation] update PR object after push

lbarcziova:
Since the PR object stores the head commit value, after the new commit push we get the statuses from the old head commit, therefore the PR object needs to be updated after the push.


csomh (approved):
Aha! So this was the issue! Thanks for fixing it :)

PR #229, opened 2021-07-23
[validation] monitor check runs instead of commit statuses

lbarcziova:
On Monday, the check runs should land in production, so this should be then merged and the image rebuilt, so that the script monitors check runs and not commit statuses.
I lowered some time limits, for submitting and completng of Copr build.

### DIFF
--- a/cron-jobs/validation/packit-service-validation.py
+++ b/cron-jobs/validation/packit-service-validation.py
@@ -14,7 +14,7 @@ from github.Commit import Commit
 
 from ogr.services.github import GithubService
 from ogr.abstract import PullRequest
-
+from ogr.services.github.check_run import GithubCheckRunStatus, GithubCheckRunResult
 
 copr = Client.create_from_config_file()
 service = GithubService(token=getenv("GITHUB_TOKEN"))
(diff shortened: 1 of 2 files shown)


cron-jobs/validation/packit-service-validation.py
```diff
@@ -141,56 +141,56 @@ def run_checks(self):
...
         :return:
         """
-        statuses = [
-            status.context
-            for status in self.get_statuses()
-            if "packit-stg" not in status.context
+        check_runs = [
+            check_run.name
+            for check_run in project.get_check_runs(commit_sha=self.head_commit)
+            if "packit-stg" not in check_run.name
```
lachmanfrantisek, line 156: It's a different story but what about having this configurable and run the validation for the stage as well so we can find the problems sooner?

lachmanfrantisek (approved):

jpopelka (approved):

cron-jobs/validation/packit-service-validation.py
lbarcziova: good idea, will include it in another PR

PR #237, opened 2021-08-04
Add variables to tweak deployment for stream service

jpopelka:


csomh:
Shouldn't the `with_repository_cache` variable also be used to exclude the repo-cache related pieces from the worker template? Or can that be done independently?

jpopelka:
It should, eventually, but this can be, I think, merged without that part.

csomh (approved):

PR #239, opened 2021-08-05
Add command with GitHub query

mfocko:
Add command with GitHub query to move-stable script that prints out URL
with filter for pull requests with release notes that are included in
current change of stable branches.

### DIFF
--- a/scripts/move_stable.py
+++ b/scripts/move_stable.py
@@ -5,6 +5,7 @@
 
 import click
 
+from datetime import date, datetime, timedelta
 from pathlib import Path
 import subprocess
 from typing import List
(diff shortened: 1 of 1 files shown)


scripts/move_stable.py
```diff
@@ -162,6 +163,40 @@ def move_all(ctx, remote, repo_store: str) -> None:
...
+        (
+            "Link with filter: https://github.com/pulls?q=org%3A{org}+"
+            "is%3Apr+"
+            "merged%3A{since}..{till}+"
+            "sort%3Aupdated-asc+"
+            "label%3A{label}"
+        ).format(
+            org="packit",
+            since=since.strftime("%Y-%m-%d"),
+            till=till.strftime("%Y-%m-%d"),
```
jpopelka, line 194: This prints out `2021-08-05 00:00:00`, while the query actually shows PRs merged today, i.e. later than 00:00:00

jpopelka (approved):
Nice!

scripts/move_stable.py
mfocko: Yeah, it is a „kind of“ issue. When moving `stable`, we usually include changes merged on Friday, but that doesn't mean we don't merge after moving `stable`. So I have to include also previous Friday.

GitHub search supports only date range, so the click does construct `datetime` object even if accepts only `date`.

Only the date is used, also the range in GitHub is both lower and upper-bound inclusive.

mfocko: OH! I know what you mean… The interval before the URL, right! will fix that...

PR #242, opened 2021-08-12
Friday (uh, actually Thursday) fix mix

jpopelka:


Makefile
```diff
@@ -7,8 +7,6 @@ AP := ansible-playbook -vv -c local -i localhost, -e ansible_python_interpreter=
 # https://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html#special-tags
 TAGS ?= all
 
-# Consider running 'make import-images' to avoid possible 'Error: ImagePullBackOff'
-# in case of newer, not-yet-imported images.
```
mfocko, line 11: As far as I understood the solution for easier rollback/redeploy, we just keep older images, which doesn't resolve the „not-yet-imported images“ issue.

mfocko (approved):
lgtm :)

Makefile
mfocko: OK, I will probably try on stage on Monday with one of the PRs that Laura has open, cause my thinking was that once the images get updated it has issue to pull the `:prod`, which should be still present currently. The only difference is, there are older images we could use (still would require to switch the tag in playbooks for redeploy).

It still seems like a clusterf*ck (no pun intended) to me, because it should take whatever is present in the **cluster** registry, so from the point of deployment it **does not** need to know about newer image on Quay, unless we do import it. (← Coming from the local deployment experience when it ranted about no images present when deploying.)

PR #247, opened 2021-08-30
Setup sandbox namespace conditionaly

jpopelka:
@TomasTomecek do you remember/know why we actually [set the finalizers to sandbox namespace](https://github.com/packit/deployment/blob/main/openshift/sandbox-namespace.yml.j2#L29)?

### DIFF
--- a/vars/stream/prod_template.yml
+++ b/vars/stream/prod_template.yml
@@ -35,6 +35,8 @@ with_pushgateway: false
 
 with_repository_cache: false
 
+with_sandbox: false
+
 # you can set the sandbox namespace name explicitly like this
 # sandbox_namespace: stream-prod-sandbox
 
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -20,6 +20,7 @@
     with_beat: true
     with_pushgateway: true
     with_repository_cache: true
+    with_sandbox: true
     push_dev_images: false
     image: quay.io/packit/packit-service:{{ deployment }}
     image_worker: quay.io/packit/packit-worker:{{ deployment }}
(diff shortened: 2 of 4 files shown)


TomasTomecek:
Nope.

[The docs are fairly complex, not sure I understand the use case fully](https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/#understanding-finalizers)

@csomh could know

TomasTomecek:
Since the sandbox namespace is not meant to be deleted, ever, I'd say the finalizer is there for protection based on the docs above. Just a guess.

csomh:
> Since the sandbox namespace is not meant to be deleted, ever, I'd say the finalizer is there for protection based on the docs above. Just a guess.

This is how I understand it, too.

jpopelka:
ok and how about a review? ;-) any ACK/NACK?

vars/stream/prod_template.yml
```diff
@@ -35,6 +35,8 @@ with_pushgateway: false
 
 with_repository_cache: false
 
+with_sandbox: false
```
csomh, line 38: Why are these set to `false` here? When should they be set to `true`? Wouldn't it be better to set the default to `false` instead?

jpopelka: These variable files (vars/**stream**/\*) are for the stream service (hardly) which AFAICT won't need the sandbox.
Packit service vars files (vars/**packit**/\*) don't set it, ie. use the default, i.e. `true`.

csomh: I'm blind, apologies.

csomh (approved):

lbarcziova (approved):

PR #260, opened 2021-10-01
Remove centosmsg

### LINKED ISSUE #225: Remove centosmsg
tl;dr; revert #71

[centosmsg](https://github.com/packit/packit-service-centosmsg) deployment was added when we started working on CentOS Stream.
Back then, repositories were in [git.centos.org](https://git.centos.org) Pagure, which exposes events to mqtt.git.centos.org, which the centosmsg listens to.
Since then the repositories have been moved to [Gitlab](https://gitlab.com/redhat/centos-stream) and we don't serve the git.centos.org repos anymore.

Note, we still use the centosmsg in [dist2src service](https://github.com/packit/dist-git-to-source-git), but it has [its own deployment](https://github.com/packit/dist-git-to-source-git/blob/main/deployment/roles/deploy/templates/centosmsg-dc.yml.j2).

jpopelka:
Fixes #225

After merge:

- [ ] Delete ImageStreams, DeploymentConfigs, pods and secrets from `stg` & `prod`
- [ ] Update the import-images cronjob


mfocko (approved):
There is also `centosmsg` in `scripts/move_stable.py`, not sure if you missed it or want to keep the `stable` branch going.

jpopelka:
> There is also `centosmsg` in `scripts/move_stable.py`

We use the stable/prod image in [dist-git-to-source-git](https://github.com/packit/dist-git-to-source-git/blob/main/deployment/roles/deploy/defaults/main.yml#L21)
Thanks for review!

PR #267, opened 2021-10-19
Remove no longer needed packit-worker-vol volumeClaimTemplate 

jpopelka:
Related: packit/packit-service#1234

---

N/A


TomasTomecek (approved):
nice, thanks!

we can then turn the sts into a deployment

PR #281, opened 2021-11-16
validation: use pr.comment instead of project.pr_comment

TomasTomecek:
this was deprecated and removed from ogr already

Fixes https://sentry.io/organizations/red-hat-0p/issues/2793426268

Tests still running locally for me.


lbarcziova (approved):
thanks!

PR #292, opened 2022-01-19
Pass PROJECT into worker

jpopelka:
~`SERVICE` is really a bad name for this variable.
I chose it to distinguish between 'packit service' and 'stream service', but 'service' has so many meanings.
We are in a deployment repo and the variable basically specifies which project to deploy to - so `PROJECT` is much better.
It's used when playbooks are created from Jinja templates, so nothing needs to be re-deployed or changed anywhere else.~

EDIT: No renaming, see comments below. This PR just passes `project` (ansible) variable to worker as `PROJECT` to be utilized [here](https://github.com/packit/packit-service/blob/17e56e0fa9121aeca44afd27cf1dba973c0b21b5/packit_service/worker/database.py#L122).

---

N/A

### DIFF
--- a/openshift/packit-worker.yml.j2
+++ b/openshift/packit-worker.yml.j2
@@ -60,6 +60,8 @@ spec:
         - name: packit-worker-1
           image: packit-worker:{{ deployment }}
           env:
+            - name: PROJECT
+              value: {{ project }}
             - name: DEPLOYMENT
               value: {{ deployment }}
             - name: DISTGIT_URL


jpopelka:
uh, nope, [we already have `project`](https://github.com/packit/deployment/blob/main/vars/template.yml#L9) which clashes with this. I need to figure out something better :-(

csomh:
Hm, if you ask me, I don't have any issue with calling it a "service": it runs in the cloud and does some things, so it's a service. It does some things for it's users (tests PRs, creates releases, keeps PRs in sync), so it's a service.

Eventually, you could call it "app" :slightly_smiling_face:

jpopelka:
Fair enough. I've left just the second commit, PTAL

openshift/packit-worker.yml.j2
```diff
@@ -60,6 +60,8 @@ spec:
         - name: packit-worker-1
           image: packit-worker:{{ deployment }}
           env:
+            - name: PROJECT
+              value: { { project } }
```
csomh, line 64:
```suggestion
              value: {{ project }}
```

jpopelka: whoops, thanks

csomh (approved):

jpopelka:
recheck

jpopelka:
recheck

PR #294, opened 2022-02-07
Increase memory limit for short-running worker

jpopelka:
It's been using 284M (271Mi) for last couple days.
Let's give it 256+64=320Mi


lachmanfrantisek (approved):
:+1: Thanks for checking that!

PR #295, opened 2022-02-10
Include {{ deployment }} in path_to_secrets

jpopelka:
so that it really points the dir with secrets, not to the parent dir.

(Let's see what happens in Zuul ;-) )


csomh (approved):
lgtm :slightly_smiling_face:

PR #299, opened 2022-02-15
Use Burstable instead of Guaranteed QoS for workers

jpopelka:
Since 9f1186bed all pods have had [Guaranteed Quality of Service (QoS)](https://medium.com/google-cloud/quality-of-service-class-qos-in-kubernetes-bb76a89eb2c6).
But for workers (long-running and all-tasks) Burstable QoS is more appropriate because:
- most of the time they don't need so many resources, but under pressure, they need 2-4x more
- we would survive even if the system (k8s/Openshift) started killing them in case of running out of memory (unlikely to happen, but still)

EDIT: I was also thinking about doing the same for `packit-service` DeploymentConfig and increasing its memory limit to 2Gi, because that's the value we needed last time when doing alembic migrations.
It can be done because we don't have any quota now and we wouldn't need to increase it manually next (each) time we do an alembic migration.
On the other hand, it moves the packit-service pod from Guaranteed to Burstable QoS, which means a higher probability of being killed if the node runs out of memory.
Not sure how likely that is, but given that there are AFAICT no quotas per project, it can IMO happen.
To stay on the safe side, I decided to leave it as it is now.

### DIFF
--- a/openshift/packit-service.yml.j2
+++ b/openshift/packit-service.yml.j2
@@ -78,7 +78,8 @@ spec:
             limits:
               # run_httpd.sh does 'alembic upgrade head' which might require more memory
               # If you see '/usr/bin/run_httpd.sh: line 16:   Killed     alembic upgrade head'
-              # you have to temporarily increase the limit (last time 512M was enough),
+              # you have to temporarily increase (in webUI/console) the limit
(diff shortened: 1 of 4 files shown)


lbarcziova (approved):
lgtm

PR #301, opened 2022-02-21
More improvements to changelog script

FrNecas:
- fix of typo from the previous PR causing a traceback
- better handling of newlines
- ignoring of pre-commit autoupdate messages
- ref no longer needs to be specified, the last tag is used by default
- README describing how to use it

### DIFF
--- a/scripts/changelog.py
+++ b/scripts/changelog.py
@@ -12,6 +12,7 @@ from git import Commit, Repo
 NOT_IMPORTANT_VALUES = ["n/a", "none", "none.", ""]
 RELEASE_NOTES_TAG = "RELEASE NOTES"
 RELEASE_NOTES_RE = f"{RELEASE_NOTES_TAG} BEGIN\n(.+)\n{RELEASE_NOTES_TAG} END"
+PRE_COMMIT_CI_MESSAGE = "pre-commit autoupdate"
 
 
 def get_relevant_commits(repository: Repo, ref: str) -> List[Commit]:
@@ -51,6 +52,8 @@ def convert_message(message: str) -> Optional[str]:
 def get_changelog(commits: List[Commit]) -> str:
     changelog = ""
     for commit in commits:
+        if PRE_COMMIT_CI_MESSAGE in commit.message:
+            continue
         message = convert_message(commit.message)
         if message and message.lower() not in NOT_IMPORTANT_VALUES:
             suffix = get_pr_data(commit.message)
--- a/scripts/changelog.py
+++ b/scripts/changelog.py
@@ -52,7 +52,7 @@ def get_changelog(commits: List[Commit]) -> str:
     changelog = ""
     for commit in commits:
         message = convert_message(commit.message)
-        if messsage and message.lower() not in NOT_IMPORTANT_VALUES:
+        if message and message.lower() not in NOT_IMPORTANT_VALUES:
             suffix = get_pr_data(commit.message)
             changelog += f"- {message} ({suffix})\n"
     return changelog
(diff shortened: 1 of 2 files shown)


scripts/changelog.py
```diff
@@ -9,9 +9,10 @@
 import click
 from git import Commit, Repo
 
-NOT_IMPORTANT_VALUES = ["n/a", "none", "none.", ""]
+NOT_IMPORTANT_VALUES = ["n/a", "none", "none.", "", "todo", "tbd"]
 RELEASE_NOTES_TAG = "RELEASE NOTES"
-RELEASE_NOTES_RE = f"{RELEASE_NOTES_TAG} BEGIN(.+){RELEASE_NOTES_TAG} END"
+RELEASE_NOTES_RE = f"{RELEASE_NOTES_TAG} BEGIN\n(.+)\n{RELEASE_NOTES_TAG} END"
```
TomasTomecek, line 14: nice! I thought that `.strip` would solve it but apparently, I haven't used it

TomasTomecek (approved):
very nice updates Franto!

FrNecas:
Will update the PR a bit more (got one idea and I will also add a README with some description to link it from our cards later)

FrNecas:
Changes done, `ref` no longer needs to be specified manually for the most common use-case :)

TomasTomecek (approved):

scripts/changelog.py
```diff
@@ -9,12 +9,21 @@
 import click
 from git import Commit, Repo
 
-NOT_IMPORTANT_VALUES = ["n/a", "none", "none.", ""]
+NOT_IMPORTANT_VALUES = ["n/a", "none", "none.", "", "todo", "tbd"]
```
lachmanfrantisek, line 12: I am not sure if we can freely skip those -- it looks like the contributor forgot but wanted to provide some changelog.

scripts/changelog.py
```diff
@@ -63,10 +74,14 @@ def get_changelog(commits: List[Commit]) -> str:
...
-    and get the changelog entry from the commit message.
-    By now, we parse a last paragraph of the pull-request description
-    (that is contained in the commit message).
-    In the future, we will have an explicit divider.
+    and gets the changelog entry from the commit message.
+    In case no REF is specified, the latest tag is used.
+
+    There are 2 possible ways to detect a changelog entry in the message:\n
+     1) The new way, beginning and end of the entry is explicitly marked
+        using `RELEASE NOTES {BEGIN,END}`.\n
```
lachmanfrantisek, line 82: I would be explicit here.
```suggestion
        using `RELEASE NOTES BEGIN` and `RELEASE NOTES END` separators.\n
```

lachmanfrantisek (approved):
Thanks!

(Nice that the last tag is now used by default.)

scripts/changelog.py
FrNecas: Hm, good point, lets not skip these. I thought about some warnings about these but there is probably no point in doing so for now. If there is a todo, the formatted output will be ` - todo (<pr number>)` which should be sufficient to track down what the PR was trying to accomplish and add the missing entry manually.

lachmanfrantisek: Yes, exactly!

jpopelka:
I haven't figured out yet what's wrong with the tests in Openshift in Zuul, but this PR doesn't touch any playbooks so go ahead and merge it manually once you're ready. You have my ACK.

FrNecas:
Dropped the timeout changes, merging manually, hopefully we can figure out why the tests fail

PR #303, opened 2022-02-23
Remove -1 suffix from packit-worker container name

jpopelka:
From what I can see in history it's been always like that.
But it's confusing since the pod names are also numbered, so we have a pod named `packit-worker-0`, which has a container named `packit-worker-1`.

For the other commits see #310

### DIFF
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -411,6 +411,14 @@
         - pushgateway
       when: with_pushgateway
 
+    - name: Wait for deploymentconfig rollouts to complete
+      # timeout 10min to not wait indefinitely in case of a problem
+      command: timeout 10m oc rollout status -w dc/{{ item }}
+      register: oc_rollout_status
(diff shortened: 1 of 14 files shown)


playbooks/deploy.yml
```diff
@@ -155,6 +155,18 @@
...
+    # can't handle when it's being populated during their startup.
+    - name: Deploy packit-worker image stream
+      k8s:
+        namespace: "{{ project }}"
+        definition: "{{ lookup('template', '{{ project_dir }}/openshift/packit-worker-is.yml.j2')  }}"
+        host: "{{ host }}"
+        api_key: "{{ api_key }}"
+        validate_certs: "{{ validate_certs }}"
+      tags:
+        - packit-worker
```
csomh, line 168: And 'beat', too, as it's using the worker image stream.

jpopelka:
shallnot

PR #311, opened 2022-03-09
Give service more memory

jpopelka:
After [moving service to F35 & mod_wsgi-express](https://github.com/packit/packit-service/pull/1363) we were seeing some errors after the pod was running for some time.
As a workaround, we [restart the process(es) periodically](https://github.com/packit/packit-service/pull/1381), which helped on stg, but not on prod.
The culprit seems to be that now we need more memory. (Kudos to @TomasTomecek for spotting it.)
From metrics, the top was 247Mi so far, so I'm setting new requests/limits to 320Mi.

Note: I'll keep an eye on the metrics and merge this next week.


TomasTomecek (commented):
seems this has helped - no issues today

lachmanfrantisek (approved):

PR #313, opened 2022-03-11
Move apiVersion: batch/v1beta1 to batch/v1 in cron jobs

jpopelka:
`batch/v1` has been available since k8s v1.21 (Openshift 4.8)
`batch/v1beta1` will no longer be served in v1.25

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125


jpopelka:
recheck

TomasTomecek (approved):

TomasTomecek:
merging this means that we can no longer test/deploy on ocp v3, right?

jpopelka:
Yes, but we don't deploy those in tests.

PR #322, opened 2022-04-22
Auto-stage cluster & ansible-lint v6

jpopelka:
This is a mix of several refactoring-ish changes, I don't think it makes sense to split them into several PRs.

I know it's risky to approve when we don't have working tests at the moment, but you have to trust me that I tested on all our projects in the new auto-stage cluster ;-)

### DIFF
--- a/playbooks/check.yml
+++ b/playbooks/check.yml
@@ -11,7 +11,7 @@
     - include_tasks: tasks/project-dir.yml
 
     - name: include variables
-      include_vars: "{{ project_dir }}/vars/{{ service }}/{{ deployment }}.yml"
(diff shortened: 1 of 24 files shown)


playbooks/check.yml
```diff
@@ -11,7 +11,7 @@
     - include_tasks: tasks/project-dir.yml
 
     - name: include variables
-      include_vars: "{{ project_dir }}/vars/{{ service }}/{{ deployment }}.yml"
+      ansible.builtin.include_vars: "{{ project_dir }}/vars/{{ service }}/{{ deployment }}.yml"
```
mfocko, line 14: „explicit is better than implicit“ to the moon… yikes

mfocko (approved):
*wink wink*

PR #324, opened 2022-04-29
fix: ignore comments in script for python dependencies

mfocko:
Strip away comment if it is put with the dependency name.

Caused by packit/ogr#710

### DIFF
--- a/scripts/setupcfg2rpm.py
+++ b/scripts/setupcfg2rpm.py
@@ -24,6 +24,7 @@ import configparser
 
 def normalize_name(name: str) -> str:
     """https://www.python.org/dev/peps/pep-0503/#normalized-names"""
+    name = name.split(";")[0]  # in case there is a comment
     return re.sub(r"[-_.]+", "-", name).lower()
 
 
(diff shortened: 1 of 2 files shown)


scripts/setupcfg2rpm.py
```diff
@@ -24,6 +24,7 @@
 
 def normalize_name(name: str) -> str:
     """https://www.python.org/dev/peps/pep-0503/#normalized-names"""
+    name = name.split(";")[0]  # in case there is a comment
```
csomh, line 27: This is a requirement marker, rather than a comment. Looking at the usage in ogr's `setup.cfg`, the dependency shouldn't be installed based on the marker in most of our images, as they have Python > "3.8".

So maybe we should evaluate these, like [this](https://github.com/csomh/deployment/commit/d62743401c8506e7004e6ffd73e7d026bd7a370d), but I'm not sure, if this is too much. This solution would also make `python3-packaging` to be a dependency for `setupcfg2rpm`.

csomh (approved):

PR #330, opened 2022-05-11
Differentiate tests by deployment

### LINKED ISSUE #318: packit-service-validation: check both stg and prod
With the more granular split of prod and stg, we can now better pinpoint jobs and `/packit[-stg]` comments to specific deployments.

At the same time, the validation script did not account for the stg deployment.

This issue tracks work so that the validation script can check both prod and stg (design and implementation are open).

This will give us daily update about the status of the stg deployment - we can be aware of problems in stg faster.

We should be able to define which test cases are triggered against both deployments to not overload the infra.

Please comment if you disagree or have ideas about stg validation.

Refinement notes:
- create a new cron job
- make the cron job configurable

majamassarini:
Fixes #318

### DIFF
--- a/cron-jobs/packit-service-validation/deploy-validation.yaml
+++ b/cron-jobs/packit-service-validation/deploy-validation.yaml
@@ -12,6 +12,8 @@
     # OR via Openshift web GUI: click on your login in top right corner, 'Copy Login Command', take the part after --token=
     api_key: ""
     validate_certs: false
+    deployment: "{{ lookup('env', 'DEPLOYMENT') }}"
+    cronjob_name: packit-{{ deployment }}-validation
 
   tasks:
     - name: Deploy secrets
@@ -31,7 +33,7 @@
     - name: Deploy cron job for running validation script
       k8s:
         namespace: "{{ project }}"
-        definition: "{{ lookup('template', './openshift/job-run-validation.yaml')  }}"
+        definition: "{{ lookup('template', './openshift/job-run-validation.yml.j2')  }}"
         host: "{{ host }}"
         api_key: "{{ api_key }}"
         validate_certs: "{{ validate_certs }}"
--- a/cron-jobs/packit-service-validation/packit-service-validation.py
+++ b/cron-jobs/packit-service-validation/packit-service-validation.py
@@ -8,6 +8,7 @@ import logging
 from copr.v3 import Client
 from datetime import datetime, timedelta, date
 from os import getenv
+from dataclasses import dataclass
 
 from github import InputGitAuthor
 from github.Commit import Commit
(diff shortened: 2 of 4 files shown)


cron-jobs/packit-service-validation/packit-service-validation.py
```diff
@@ -378,21 +441,31 @@ def check_comment(self):
...
         logging.warning("SENTRY_SECRET was not set!")
 
+    if getenv("DEPLOYMENT"):
+        deployment = (
+            Staging() if getenv("DEPLOYMENT") == Deployments.staging else Production()
+        )
+    else:
+        deployment = Staging()
+        # deployment = Production()
+
```
lbarcziova, line 451:
```suggestion
deployment = Production() 
                      if getenv("DEPLOYMENT", Deployments.production) == Deployments.production
                      else Staging()
```
I am also thinking about whether the default should be staging or production :thinking: but the `Testcase` class uses production as default

cron-jobs/packit-service-validation/packit-service-validation.py
```diff
@@ -17,29 +18,68 @@
...
-APP_NAME = "Packit-as-a-Service"
-
 copr = Client({"copr_url": "https://copr.fedorainfracloud.org"})
 service = GithubService(token=getenv("GITHUB_TOKEN"))
 project = service.get_project(repo="hello-world", namespace="packit")
 user = InputGitAuthor(name="Release Bot", email="user-cont-team+release-bot@redhat.com")
 logging.basicConfig(level=logging.WARNING)
 
 
+class Deployments(str, enum.Enum):
```
lbarcziova, line 28: I would probably go with singular, but it's maybe just personal preference

lbarcziova (commented):
very nice changes!

cron-jobs/packit-service-validation/packit-service-validation.py
majamassarini: I also prefer set `Production` as default, you were such speedy in these review that I had not time to fix it from my test version ^_^

cron-jobs/packit-service-validation/packit-service-validation.py
majamassarini: Usually I also prefer the singular but in these case then it clashes with the base class which holds the Deployment info...
So I don't know what could be better... any suggestion welcome!

cron-jobs/packit-service-validation/deploy-validation.yaml
```diff
@@ -31,7 +31,7 @@
     - name: Deploy cron job for running validation script
       k8s:
         namespace: "{{ project }}"
-        definition: "{{ lookup('template', './openshift/job-run-validation.yaml')  }}"
+        definition: "{{ lookup('template', './openshift/job-run-validation.yaml', './openshift/job-run-validation-stg.yaml')  }}"
```
lbarcziova, line 34: could we use the existing template and use a variable (e.g.`deployment` as [here](https://github.com/packit/deployment/blob/ac058d2e663e8e7d6886be1b9547a65796e02d68/playbooks/deploy.yml#L11)) and based on that variable cron job `name`  and `DEPLOYMENT` env var will be adjusted?

cron-jobs/packit-service-validation/packit-service-validation.py
lbarcziova: ah, I haven't realised that...If the other class holds only information, it could be somehow reflected in the name, e.g. DeploymentVariables/DeploymentInfo, but I am not the expert in naming, so I will leave it up to you and I am okay with having it as it now :D

cron-jobs/packit-service-validation/deploy-validation.yaml
majamassarini: Sure, I have miss-understood what `lookup` is doing 😅

mfocko (commented):
lgtm :)

cron-jobs/packit-service-validation/openshift/job-run-validation-stg.yaml
```diff
@@ -0,0 +1,40 @@
+# Copyright Contributors to the Packit project.
```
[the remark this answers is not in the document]
majamassarini: I can try!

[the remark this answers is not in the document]
lbarcziova: I am personally more inclined to have it running in Openshift and independent of Github

PR #334, opened 2022-05-13
Merge pull request #330 from majamassarini/fix/deployment/318

majamassarini:
Dataclass object are not like expected using dataclass inheritance.
We can use typing.Union to define a type which refers both dataclasses.


TomasTomecek (approved):
I have never used dataclasses myself, so I played with this in an interactive python shell and it worked as expected

PR #337, opened 2022-05-17
Push image to quay using commit SHA as tag

majamassarini:
To be able to use the Unified Openshift Deployment process we need to push images to Quay using a commit SHA tag


.github/workflows/build-and-push-cronjob-image.yaml
```diff
@@ -47,6 +47,7 @@ jobs:
           image: ${{ matrix.image }}
           context: ${{ env.path }}
           oci: true
+          tags: latest "${ GITHUB_SHA::7 }"
```
jpopelka, line 50: I think it's either `${GITHUB_SHA::7}` (without the whitespaces) or `${{ GITHUB_SHA::7 }}` (but that's I think syntax for [expressions](https://docs.github.com/en/actions/learn-github-actions/expressions) or [contexts](https://docs.github.com/en/actions/learn-github-actions/contexts)).
And I'd probably try to remove the quotes as well.
```suggestion
          tags: latest ${GITHUB_SHA::7}
```

jpopelka (approved):
OK, let's try.

jpopelka:
Just merge manually. From what I remember this happens when one touches the github workflow.

PR #338, opened 2022-05-18
Describe how to add a new package to be served by staging stream bot

jpopelka:


majamassarini (approved):
LGTM

I have just one dummy question: instead of cloud-init I could use python-teamcity-messages without asking further permissions?

jpopelka:
> I could use python-teamcity-messages without asking further permissions?

I'm afraid you can't because CentOS Stream (RHEL) contains only subset of Fedora packages and as you can check in the [CentOS Stream dist-git](https://gitlab.com/redhat/centos-stream/rpms) `python-teamcity-messages` in not one of those.

I've just invited you to https://gitlab.com/packit-service so that you can play with the source-git repos in https://gitlab.com/packit-service/src/ (I'll usually open a PR against https://gitlab.com/packit-service/src/open-vm-tools or https://gitlab.com/packit-service/src/luksmeta)

PR #351, opened 2022-06-22
Celery_metrics

majamassarini:
Use flower service to collect celery related metrics (https://issues.redhat.com/browse/PACKIT-1109).

Related with packit/packit-service#1561

### DIFF
--- a/openshift/nginx.yml.j2
+++ b/openshift/nginx.yml.j2
@@ -25,7 +25,7 @@ spec:
           secretName: packit-secrets
       - name: flower-secret
         secret:
-          secretName: flower-secret
+          secretName: flower-htpasswd
       containers:
       - name: nginx
         image: nginxinc/nginx-unprivileged
@@ -77,6 +77,7 @@ data:
                proxy_method     GET;
              }
 
+             # https://flower.readthedocs.io/en/1.0/reverse-proxy.html
              location / {
                proxy_pass http://flower:5555;
                proxy_set_header Host $host;
@@ -87,7 +88,7 @@ data:
                proxy_set_header Connection "upgrade";
 
                auth_basic  "Restricted";
-               auth_basic_user_file  /etc/nginx/secrets/flower-basic-auth;
+               auth_basic_user_file  "/etc/nginx/secrets/htpasswd";
              }
 
          }
--- a/openshift/nginx.yml.j2
+++ b/openshift/nginx.yml.j2
@@ -23,6 +23,9 @@ spec:
       - name: packit-secrets
         secret:
           secretName: packit-secrets
+      - name: flower-secret
+        secret:
+          secretName: flower-secret
       containers:
       - name: nginx
         image: nginxinc/nginx-unprivileged
@@ -33,6 +36,8 @@ spec:
           mountPath: /etc/nginx
         - name: packit-secrets
           mountPath: /secrets
+        - name: flower-secret
+          mountPath: /etc/nginx/secrets
         resources:
           # requests and limits have to be the same to have Guaranteed QoS
           requests:
@@ -67,9 +72,24 @@ data:
                 proxy_method     GET;
              }
 
-             location / {
-                return           404;
+             location = /flower-metrics {
+               proxy_pass http://flower:5555/metrics;
+               proxy_method     GET;
              }
+
+             location / {
+               proxy_pass http://flower:5555;
+               proxy_set_header Host $host;
+               proxy_set_header X-Real-IP $remote_addr;
+
+               proxy_http_version 1.1;
+               proxy_set_header Upgrade $http_upgrade;
+               proxy_set_header Connection "upgrade";
+
+               auth_basic  "Restricted";
+               auth_basic_user_file  /etc/nginx/secrets/flower-basic-auth;
+             }
+
          }
     }
 ---
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -359,13 +359,31 @@
         - pushgateway
       when: with_pushgateway
 
-    - name: Deploy nginx to reverse proxy the pushgateway
+    - name: Create htpasswd file with flower-boss passwd
+      htpasswd:
+        path: "{{ flower_secret_htpasswd }}"
+        name: "flower-boss"
+        password: "{{ flower_basic_auth | regex_replace('flower-boss:', '') }}"
+        mode: 0640
+      tags:
+        - flower
+      when: with_flower
+
+    - name: Upload flower secret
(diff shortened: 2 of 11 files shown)


TomasTomecek (approved):

TomasTomecek:
Where do we install the tool that provides the prometheus metrics? Or is this supported out of the box by flower?

openshift/nginx.yml.j2
```diff
@@ -67,9 +72,24 @@ data:
                 proxy_method     GET;
              }
 
+             location = /flower-metrics {
+               proxy_pass http://flower:5555/metrics;
+               proxy_method     GET;
+             }
+
              location / {
```
csomh, line 80:
```suggestion
             location /flower {
```

I think this should be rather something like this, in order to be explicit about where this route is going to lead :)

openshift/nginx.yml.j2
```diff
@@ -67,9 +72,24 @@ data:
...
-                return           404;
+               proxy_pass http://flower:5555;
+               proxy_set_header Host $host;
+               proxy_set_header X-Real-IP $remote_addr;
+
+               proxy_http_version 1.1;
+               proxy_set_header Upgrade $http_upgrade;
+               proxy_set_header Connection "upgrade";
```
csomh, line 87: Why are these required? Maybe add a comment or link explaining it.

playbooks/deploy.yml
```diff
@@ -359,13 +360,47 @@
...
+
+    - name: Create htpasswd file with flower-boss passwd
+      htpasswd:
+        path: "{{ flower_secret_htpasswd }}"
+        name: "flower-boss"
+        password: "{{ flower_basic_auth | regex_replace('flower-boss:', '') }}"
+        mode: 0640
+      tags:
+        - flower
+      when: with_flower
```
csomh, line 383: Seeing it like this, I think it would be much easier to not to generate this file on every Ansible playbook run, but to generate it manually and save it as an attachment in Bitwarden. Then it could be deployed with the rest of the secrets.

csomh: And we wouldn't need to introduce new dependencies :)

openshift/nginx.yml.j2
csomh: And keep returning `404` on `/`.

majamassarini:
Flower has already all the metrics provided by celery-exporter, for this reason we decided (with Laura) to use it.
This changes are needed to make the flower metrics be reachable (by prometheus) without any authentication and let all the other flower services be under authentication as they already were.

openshift/nginx.yml.j2
majamassarini: I tried really hard to do that, without any success.
I was not able to serve the static content as expected by flower and at the end I give up... if you think this is necessary we can try to configure it together.

playbooks/deploy.yml
majamassarini: We need both the plain password (to access flower services manually) and the crypted password (to generate the htpasswd like file for nginx).
If we do it by hand we should cope with the "duplication" of having the same password in two different formats.
I am not really sure that this is a good idea but, on the other hand, I am not sure that my approach is the easiest and cleanest, maybe do we have some other way to do that?

openshift/nginx.yml.j2
csomh: After playing with this a little bit, I don't think this can be done, or at least I wasn't able to figure it out either. Flower needs some more resources (CSS, JS) loaded to run, and loading these resources will happen using the hostname, but in order to work, it should use `hostname/flower`. And I wasn't able to figure out how to configure this in flower.

So let's keep everything as it is.

playbooks/deploy.yml
csomh: I don't mind the duplication in this case, as it will help reduce complexity.

playbooks/deploy.yml
```diff
@@ -359,13 +360,47 @@
...
+        path: "{{ flower_secret_htpasswd }}"
+        state: absent
+      tags:
+        - flower
+      when: with_flower
+
+    - name: Show flower-boss passwd
+      ansible.builtin.debug:
+        msg: "{{ flower_basic_auth | regex_replace('flower-boss:', '') }}"
+
```
csomh, line 374:
```suggestion
```

Could be removed.

playbooks/deploy.yml
```diff
@@ -359,13 +360,47 @@
...
+    - name: Create htpasswd file with flower-boss passwd
+      htpasswd:
```
csomh, line 376:
```suggestion
      community.general.htpasswd:
```

playbooks/deploy.yml
```diff
@@ -359,13 +360,47 @@
...
 
-    - name: Deploy nginx to reverse proxy the pushgateway
+    - name: Delete previous created file with flower credentials for nginx, if any
+      ansible.builtin.file:
+        path: "{{ flower_secret_htpasswd }}"
+        state: absent
+      tags:
+        - flower
+      when: with_flower
+
```
csomh, line 370: There is no need to do this, as the `htpasswd` module will update the file if any of the credentials do change.

playbooks/deploy.yml
jpopelka: Both solutions have pros and cons. First I leaned toward Hunor's suggestion to have "duplicated" file(s) in the vault rather than more code here in the playbook. But then, we would need to document somewhere how we actually created that file, because I definitely won't remember after some time. So now I'm +1 for creating the file on the fly as it's just a few simple lines here in the playbook.

vars/packit/prod_template.yml
```diff
@@ -31,7 +31,7 @@ api_key: ""
 
 # with_redis_commander: false
 
-# with_flower: false
+with_flower: true
```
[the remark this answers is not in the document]
jpopelka: forget about it :-) the main template.yml is fine since we're not changing the default value of `with_flower`

playbooks/deploy.yml
```diff
@@ -28,6 +28,7 @@
     image_tokman: quay.io/packit/tokman:{{ deployment }}
     # project_dir is set in tasks/project-dir.yml
     path_to_secrets: "{{ project_dir }}/secrets/{{ service }}/{{ deployment }}"
+    flower_secret_htpasswd: "{{ path_to_secrets }}/flower-htpasswd"
```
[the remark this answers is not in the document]
majamassarini: Nice, thanks for explaining me this!

csomh (approved):

PR #352, opened 2022-06-22
Add fluentd sidecar to worker

jpopelka:
https://github.com/packit/research/pull/152

We use our image https://quay.io/repository/packit/fluentd-splunk-hec, built via [a GitHub workflow](https://github.com/jpopelka/fluent-plugin-splunk-hec/blob/main/.github/workflows/rebuild-and-push-image.yml) because we don't want to use [docker.io/splunk/fluentd-hec image](https://hub.docker.com/r/splunk/fluentd-hec).

I've been unsuccessful in sending data to Splunk HEC so far.
This is a middle step to make sure the fluentd sidecar works OK and at this moment we're [sending logs to files](https://docs.fluentd.org/output/file) in a persistent volume mount to `/var/log/packit/` in the fluentd container.

Once a day is the intermediate `buffer.*.log` file in `/var/log/packit/` flushed and compressed into a permanent gzip file.
The file is `.` prefixed, so use `ls -a` to see it.

The `fluentd-config` has also a match section with `@type splunk_hec`, but it's not matched :) now because the preceding `@type file` is matched first.

The fluentd sidecar has no resource requests/limits because with it the fluentd was starting very slow (about 10sec) and it missed some logs from worker which was already up and processing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ This file documents basic usage, for more info see
   - [Packit service deployment specifics](docs/packit-service.md)
   - [Fedora source-git bot deployment specifics](docs/fedora-source-git.md)
   - [CentOS Stream source-git bot deployment specifics](docs/centos-stream-source-git.md)
+  - [Logs](docs/logs.md) - how we aggregate worker logs and where to find them
 - [playbooks/](playbooks/) - Ansible playbooks
 - [roles/](roles/) - Ansible roles
 - [vars/](vars/) - Variable file(s). See [vars/README.md](vars/README.md).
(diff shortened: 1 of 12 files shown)


jpopelka:
Ready for review, it's been running in stg since yesterday.

TomasTomecek (changes requested):
Even though I think I know what this change does, I'd appreciate a brief explanation in a README: explain what fluentd is, why we use it, what it does in our deployment, how & where are the logs stored; thanks!

Implementation LGTM

jpopelka:
> I'd appreciate a brief explanation in a README

Added [docs/logs.md](https://github.com/packit/deployment/blob/f011d68f8b1972303362dacc117abef3e010a64a/docs/logs.md)

TomasTomecek (approved):
👋👋👋

very nice write-up, thank you!

PR #364, opened 2022-08-05
Move stable branches also for ogr and specfile

nforro:


TomasTomecek (approved):
I verified both stable branches are set up, thanks!

PR #367, opened 2022-08-16
Add community.general Ansible collection

nforro:
To be able to easily enable/disable copr repos.


jpopelka (approved):
I'd like to keep the base image as small as possible so I was wondering whether it makes sense to install it here if we could just do `dnf copr enable user/project` in the playbooks.

OTOH, once we move the base image to F36 the main `ansible` package will take twice as much space because it already contains the community plugins, so I guess it's OK.

Maybe, if you could just put a comment there saying to remove the `ansible-collection-community-general` when moving to F36?

nforro:
> Maybe, if you could just put a comment there saying to remove the `ansible-collection-community-general` when moving to F36?

Ok, I'll do that.

PR #372, opened 2022-08-25
Autoscale prod short-running worker & make stg workers layout similar to prod

jpopelka:
I'll merge once this proves to work on prod. Let's discuss it meanwhile.

### DIFF
--- a/openshift/packit-worker.yml.j2
+++ b/openshift/packit-worker.yml.j2
@@ -115,6 +115,12 @@ spec:
                 secretKeyRef:
                   name: aws
                   key: secret-access-key
+{% if queues == 'short-running' %}
+            - name: AUTOSCALE
+              value: "8,1"
+            - name: POOL
+              value: gevent
+{% endif %}
           volumeMounts:
             - name: packit-ssh
               mountPath: /packit-ssh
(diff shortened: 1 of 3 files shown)


openshift/packit-worker.yml.j2
```diff
@@ -115,6 +115,12 @@ spec:
                 secretKeyRef:
                   name: aws
                   key: secret-access-key
+{% if queues == 'short-running' %}
+            - name: AUTOSCALE
+              value: "8,1"
+            - name: POOL
+              value: gevent
```
csomh, line 122: Apologies for my ignorance: how will these values cause autoscaling to happen? What does `"8,1"` mean?

jpopelka: https://github.com/packit/research/tree/main/workers-scaling#worker---autoscale
https://github.com/packit/packit-service/blob/main/files/run_worker.sh#L33

Should I put some comment and/or link into the `openshift/packit-worker.yml.j2`?

csomh: Nope, there is no need. I should have grepped for these names...

lachmanfrantisek (approved):

PR #373, opened 2022-09-01
Give packit-prod postgres 2Gi memory

jpopelka:
It a bit worries me that it's been just a couple of months since I last increased it from 512Mi to 1Gi (898c2ea5c) and it's again not enough.
We can afford it so let's do it, but next time we should probably play with `POSTGRESQL_SHARED_BUFFERS` & `POSTGRESQL_EFFECTIVE_CACHE_SIZE` whose [default values are based on the memory limit](https://github.com/sclorg/postgresql-container/blob/master/src/root/usr/share/container-scripts/postgresql/common.sh#L12), so the more memory we give it the bigger buffers/cache it uses.


jpopelka:
Screenshot from 2022-09-01 13-15-10

lachmanfrantisek (approved):

majamassarini (approved):

TomasTomecek (approved):

PR #376, opened 2022-09-12
Use Deployments instead of DeploymentConfigs

jpopelka:
OpenShift introduced `DeploymentConfig` objects before `Deployments` were added to upstream Kubernetes.
[Deployments are now in general preferred in OCP 4](https://docs.openshift.com/container-platform/4.11/applications/deployments/what-deployments-are.html#deployments-comparing-deploymentconfigs_what-deployments-are).

For me personally, the benefit would be that:
- I could see Metrics for the Deployment without having to go to Pods first
- the *-deploy pods would no longer (as I understand it) clutter the list of all (project's) pods

### DIFF
--- a/docs/continuous-deployment.md
+++ b/docs/continuous-deployment.md
@@ -8,9 +8,9 @@ as intermediary between an image registry (Quay.io) and a Deployment/StatefulSet
 It has several significant benefits:
 
 - We can automatically trigger Deployment when a new image is pushed to the registry.
-- We can rollback/revert/undo the Deployment.
+- We can roll back/revert/undo the Deployment.
 
-`Image registry` -> [1] -> `ImageStream` -> `DeploymentConfig`/`StatefulSet`
+`Image registry` -> [1] -> `ImageStream` -> `Deployment`/`StatefulSet`
 
 [1] This is automatic (even it can take some time) on stg, but not on prod.
 On prod, where we don't want the images to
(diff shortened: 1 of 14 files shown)


TomasTomecek:
Agreed, this is a welcomed change :)

those merge conflicts...

TomasTomecek (commented):
So deployment doesn't have any abbreviation? That's a downgrade to type deployment instead of dc :D although I don't interact with that object much so it should be okay

Will you delete all DCs and then run `make deploy` when rolling out this change?

jpopelka:
> That's a downgrade to type deployment instead of dc

I feel your pain, bro ;)

> Will you delete all DCs and then run `make deploy` when rolling out this change?

probably

Now, I'm just worried that the `oc cluster up` (Openshift v3) in Zuul doesn't know the `Deployment` so I'll need to somehow template this, I tried to find when was the `Deployment` added to k8s/Openshift, but didn't find it.

TomasTomecek (approved):

openshift/packit-service-beat.yml.j2
```diff
@@ -2,19 +2,22 @@
...
+apiVersion: apps/v1
 metadata:
-  labels:
-    app: packit
-    name: packit-service-beat
   name: packit-service-beat
+  annotations:
+    # https://docs.openshift.com/container-platform/4.11/openshift_images/triggering-updates-on-imagestream-changes.html
+    image.openshift.io/triggers: >-
+      [{"from":{"kind":"ImageStreamTag","name":"packit-worker:{{ deployment }}"},"fieldPath":"spec.template.spec.containers[?(@.name==\"packit-service-beat\")].image"}]
```
TomasTomecek, line 12: one would think openshift would do this more nicely

docs/continuous-deployment.md
```diff
@@ -46,13 +46,13 @@ There's also 'import-images' target in the Makefile, so `DEPLOYMENT=prod make im
 
 ### Reverting to older deployment/revision/image
 
-`DeploymentConfig`s can be reverted with `oc rollout undo`:
+`Deployment`s can be reverted with `oc rollout undo`:
 
-    $ oc rollout undo dc/packit-service [--to-revision=X]
-    $ oc rollout undo dc/packit-service-fedmsg [--to-revision=X]
+    $ oc rollout undo deploy/packit-service [--to-revision=X]
+    $ oc rollout undo deploy/packit-service-fedmsg [--to-revision=X]
```
TomasTomecek, line 52: 4 more letters is not that bad :)

TomasTomecek (approved):

openshift/packit-service-beat.yml.j2
jpopelka: OpenShift does - with `DeploymentConfig`, which is, along with `ImageStream`, OpenShift's resource type.
But `Deployment` is k8s resource type and by itself doesn't know how to work with `ImageStream`s.

jpopelka:
Everything looks OK on staging, merging.

PR #387, opened 2022-10-18
Move non-secret configurations from vault to this repo

### LINKED ISSUE #368: Store the non-secret Packit service configuration options in vars/
Currently, the Packit service configuration contains both secrets and real configuration options. This causes that each time we want to update the real configuration options, we need to interact with Bitwarden. To avoid this, store these options in this repo publicly (probably in vars/) and inject the secrets to the config during the deployment.
TODO:
- [ ] move the real configuration options from Bitwarden to `vars` in this repo
- [ ] inject the secret values into the Packit service config during the deployment

jpopelka:
Instead of storing a whole `packit-service.yaml` and `fedora.toml` in the vault, we store in the vault (in `extra-vars.yml`) only the private values.
The `packit-service.yaml.j2` and `fedora.toml.j2` templates are stored in `secrets/<service>/<deployment>` and processed during deployment.

Fixes #368 

- [ ] Where do we use [secrets/template/](https://github.com/packit/deployment/tree/main/secrets/template) ? Document it or remove it.


mfocko:
> Where do we use [secrets/template/](https://github.com/packit/deployment/tree/main/secrets/template) ? Document it or remove it.

IIRC it is used in packit-service (openshift) tests with deploying or Zuul in here

jpopelka:
> IIRC it is used in packit-service (openshift) tests with deploying or Zuul in here

I can see that [packit-service/files/deployment.yaml runs `make generate-local-secrets` in a cloned deployment repo](https://github.com/packit/packit-service/blob/main/files/deployment.yaml#L43), but that calls the [generate-secrets role](https://github.com/packit/deployment/tree/main/roles/generate-secrets) which has it's own templates.

Ditto [playbooks/generate-local-secrets.yml](https://github.com/packit/deployment/blob/main/playbooks/generate-local-secrets.yml) in this repo.

openshift/secret-aws.yml.j2
```diff
@@ -8,5 +8,5 @@ metadata:
   name: aws
 type: Opaque
 data:
-  access-key-id: "{{ aws_access_key_id | b64encode }}"
-  secret-access-key: "{{ aws_secret_access_key | b64encode }}"
+  access-key-id: "{{ vault.aws.access_key_id | b64encode }}"
+  secret-access-key: "{{ vault.aws.secret_access_key | b64encode }}"
```
TomasTomecek, line 12: this looks much more usable

secrets/README.md
```diff
@@ -1,9 +1,10 @@
...
 
-Deployment process (`make deploy`) expects files to be transformed to
-Openshift secrets to be found in these subdirectories.
-These files are either automatically downloaded (`make download-secrets`)
-or they need to be created manually in case of local/dev/test deployment.
+During deployment (`make deploy`), secret files are downloaded from a vault
+into one of the subdirectories and transformed into Kubernetes/Openshift secrets.
+They can also be downloaded independently with `make download-secrets`.
+By default, these subdirectories contain only templates, which don't contain any secret,
+but which are processes and injected with secrets during the deployment process.
```
TomasTomecek, line 7: does it mean that I need to run `make deploy` to get the templates rendered to see the data?

could we get a make target that could do the rendering?

TomasTomecek (approved):
looking good, just one comment about the usability

secrets/README.md
jpopelka: With `data` you mean just the secret values (they are in `extra-vars.yml`) or the processed (with secret values) `packit-service.yaml`?
If the latter, then yes, your only option is to check what's been deployed after you run the deployment.
But do you really need to see it processed before you run the deployment? What the use case? If there's a syntax error for example in the `authentication` dictionary it should be easier to spot in the `extra-vars.yml` than in the whole `packit-service.yaml`.

TomasTomecek: okay, that makes sense; I was just concerned that this will make the deploy process more complex (will require rendering), but at the same time, it's not a problem; as you write, if I wanna see the values, I can look into extra-vars.yaml

jpopelka: @TomasTomecek here you go #390

PR #390, opened 2022-10-20
Add `render-secrets-from-templates` target

jpopelka:
You need it if you want to create `packit-service.yaml` in order to run
`$ scripts/update_oc_secret.sh packit-config secrets/packit/stg/packit-service.yaml`

I've been thinking about whether to call it directly after downloading the secrets from the vault, but I'm leaving it separated so that people are aware that it's generated and not downloaded from the vault.

Also, it's not needed during the deployment so it'd slow it down.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,11 @@ TAGS ?= all
 download-secrets:
 	./scripts/download_secrets.sh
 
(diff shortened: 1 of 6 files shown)


TomasTomecek (approved):
😍

PR #399, opened 2022-11-09
Increase postgres resources 1->1.5Gi

jpopelka:
Last time (2f07b2318) we tried to lower the value of shared buffers but it hasn't helped much so it's probably time to request more memory. The shared buffers value is 15% of the request/limit.
Screenshot from 2022-11-08 17-54-52


jpopelka:
After [cleaning more old SRPM logs](https://github.com/packit/packit-service/pull/1745/commits/84e8f03465f8a006145f04225369444aa6effc4e) the memory consumption has lowered to 0.8GiB so this might not be needed after all. (or not now)

PR #411, opened 2022-11-23
One 16-threaded short-running worker @ prod

jpopelka:
kinda reverts 7b074a7ae but don't specify `POOL` because of packit/packit-service#1770

I'll merge next week once packit/packit-service#1770 is in stable/prod.

### DIFF
--- a/openshift/packit-worker.yml.j2
+++ b/openshift/packit-worker.yml.j2
@@ -95,6 +95,10 @@ spec:
               valueFrom: {secretKeyRef: {name: aws, key: access-key-id}}
             - name: AWS_SECRET_ACCESS_KEY
               valueFrom: {secretKeyRef: {name: aws, key: secret-access-key}}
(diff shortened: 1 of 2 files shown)


openshift/packit-worker.yml.j2
```diff
@@ -95,6 +95,10 @@ spec:
               valueFrom: {secretKeyRef: {name: aws, key: access-key-id}}
             - name: AWS_SECRET_ACCESS_KEY
               valueFrom: {secretKeyRef: {name: aws, key: secret-access-key}}
+{% if queues == 'short-running' %}
+            - name: CONCURRENCY
+              value: "16"
```
TomasTomecek, line 100: wow, 16 :O let's test those race conditions :)

jpopelka: It's been like that for quite some time already.

PR #414, opened 2022-11-29
Adjust resource limits based on prod metrics

jpopelka:
Looking at packit prod resource utilization graphs, one can see that we're requesting 1.4cpu, but utilizing 0.1-0.4cpu, so we're too greedy.

I went through packit-prod metrics for individual components for the past 2 weeks and adjusted the resource limits to be 2-3x the values at peaks so we should still be safe.

You can see that it's mostly adjusting cpu, but in some cases, it decreases the memory limit from 80Mi to 32Mi.
Those are components that need only a few MiB of memory but previously had the limit set to 80Mi because of Openshift Online where the limit couldn't be less than 80Mi, see 746cf271692

I already deployed this to `packit-stg` so you can check that everything works as expected.

### DIFF
--- a/openshift/nginx.yml.j2
+++ b/openshift/nginx.yml.j2
@@ -44,11 +44,11 @@ spec:
         resources:
           # requests and limits have to be the same to have Guaranteed QoS
           requests:
-            memory: "80Mi"
-            cpu: "50m"
+            memory: "32Mi"
+            cpu: "10m"
           limits:
-            memory: "80Mi"
-            cpu: "50m"
+            memory: "32Mi"
+            cpu: "10m"
   replicas: 1
   strategy:
     type: Recreate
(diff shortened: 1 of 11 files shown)


openshift/nginx.yml.j2
```diff
@@ -44,11 +44,11 @@ spec:
         resources:
           # requests and limits have to be the same to have Guaranteed QoS
           requests:
-            memory: "80Mi"
-            cpu: "50m"
+            memory: "32Mi"
```
TomasTomecek, line 47: nice, didn't know that ngingx is so efficient

TomasTomecek (approved):
agreed! we don't need that much CPU since we're not doing much compute

if we rewrote packit in golang, we'd even need much less memory :D /s

openshift/nginx.yml.j2
jpopelka: yeah, it's currently taking just 6MiB@prod & 3MiB@stg :-)

Issue #396, opened 2022-11-02
Replace 'set-output' in GitHub workflows with environment files


csomh:
`set-output` used in several workflows in github.com/packit is going to be deprecated on _31st May 2023_. See the [blogpost](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) explaining the change and how it should be replaced with "environment files".

PR #418, opened 2022-12-07
Use environment files instead of set-output command

jpopelka:
Related to #396

* https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
* https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

### DIFF
--- a/.github/workflows/build-and-push-cronjob-image.yml
+++ b/.github/workflows/build-and-push-cronjob-image.yml
@@ -26,13 +26,12 @@ jobs:
       path: cron-jobs/${{ matrix.image }}/
 
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
 
       - name: Add short SHA to the list of tags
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-push-cronjob-image.yml
```diff
@@ -26,13 +26,12 @@ jobs:
...
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
 
       - name: Add short SHA to the list of tags
         shell: bash
         run: |
-          echo "::set-output name=tags::latest ${GITHUB_SHA::7}"
+          echo "tags=latest ${GITHUB_SHA::7}" >> $GITHUB_OUTPUT
```
TomasTomecek, line 34: I wonder where does that env var point :)

TomasTomecek (approved):

PR #420, opened 2022-12-12
fix(import): add fluentd sidecar to image streams

mfocko:
When running `make import-images`, we were missing a fluentd sidecar image, so let's add it to the playbook :)

### DIFF
--- a/playbooks/import-images.yml
+++ b/playbooks/import-images.yml
@@ -10,6 +10,7 @@
     with_fedmsg: true
     with_dashboard: true
     with_tokman: true
+    with_fluentd_sidecar: true
   tasks:
     - name: Include variables
       ansible.builtin.include_vars: ../vars/{{ service }}/{{ deployment }}.yml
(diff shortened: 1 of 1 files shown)


playbooks/import-images.yml
```diff
@@ -19,6 +20,11 @@
       # it doesn't change anything, so don't report 'changed'
       changed_when: false
 
+    - name: Check/import newer fluentd image into image stream
+      ansible.builtin.command: oc import-image is/fluentd:{{ deployment }} -n {{ service }}-{{ deployment }}
```
TomasTomecek, line 24: I couldn't find a definition of the fluentd imagestream: do we need to do that first or is this enough?

jpopelka: The Image stream is created together with workers in
https://github.com/packit/deployment/blob/main/openshift/packit-worker.yml.j2#L192

This playbook expects the image streams already exist.

jpopelka (approved):

PR #428, opened 2023-01-04
move-stable: make specific instructions more visible

TomasTomecek:
This also prints latest builds of our packages.

### DIFF
--- a/scripts/move_stable.py
+++ b/scripts/move_stable.py
@@ -9,6 +9,7 @@ import subprocess
 from typing import List, Optional
 
 import click
+from copr.v3 import Client
 from git import Repo
 
 import changelog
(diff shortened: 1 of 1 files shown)


scripts/move_stable.py
```diff
@@ -354,22 +355,55 @@ def push_stable_branch(path_to_repository: Path, remote: str, commit_sha: str) -
...
+                "packit-service images install ogr/specfile/packit from "
+                "https://copr.fedorainfracloud.org/coprs/packit/packit-stable/builds/ "
+                "so make sure the builds are done before you proceed!\n"
+                "You can also proceed now and rebuild the stable images later in "
+                "https://github.com/packit/packit-service/actions"
+                "\n====================================================================\n",
+                fg="red",
+            )
         )
+        click.echo("Are these the latest builds?")
```
jpopelka, line 405: why a different echo in each repo?
```suggestion
        click.echo("These are the latest stable builds:")
```

jpopelka (approved):
coloooours :eyes: :smiley:

scripts/move_stable.py
TomasTomecek: My original idea was to reference the famous Star wars quote

> "These aren't the droids you're looking for."

but failed

nevertheless we spoke with Jirka in person and concluded this is not an ideal description but should be good enough: we turn it from a question to a statement if someone sends a PR

PR #431, opened 2023-01-11
Yet another update of resource requests & limits

jpopelka:
Use Burstable QoS for more/most pods, not just workers (d9278288).

Guaranteed QoS (when requests == limits) is the 'best' class, but you
have to carefully set the values (the same for requests and limits) so that
- it's enough for all use cases, but at the same time
- it's not too much because you'd waste the reserved (requested) resources.

And you actually don't need the generic workloads to be 'guaranteed'.
The Guaranteed QoS makes sense for the 'sensitive' workloads, like in
our case postgres and redis (maybe the service/api).

With the Burstable QoS (requests < limits) we can set
- requests to what the pod needs most of the time
- limits to some high value (more than what the pod needs in spikes)
  to have a safety net when a pod gets crazy
  - when it becomes a memory hog due to a memory leak
  - when it starts eating all cpu because of a bug

### DIFF
--- a/openshift/dashboard.yml.j2
+++ b/openshift/dashboard.yml.j2
@@ -32,13 +32,12 @@ spec:
             - name: DEPLOYMENT
               value: {{ deployment }}
           resources:
-            # requests and limits have to be the same to have Guaranteed QoS
             requests:
               memory: "128Mi"
               cpu: "20m"
             limits:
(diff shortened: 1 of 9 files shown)


TomasTomecek (approved):
makes sense to me, thank you for the detailed explanation

PR #432, opened 2023-01-17
fix(scripts,stable): disable progress bar for no deps

mfocko:
When there are no Copr dependencies to wait for, current implementation did:

    Waiting for Copr builds [------------------------------------] 0%

Switch it, so that for no dependencies it just prints out an information about no dependencies being configured.


lbarcziova (approved):
lgtm

PR #434, opened 2023-01-18
Set kerberos credentials/ticket cache location

jpopelka:
The `default_ccache_name = KEYRING:persistent:%{uid}` from `/etc/krb5.conf` didn't work for us after OpenShift 4.10->4.11 upgrade. We were getting `kinit: Function not implemented while getting default ccache`.

Kudos to @lachmanfrantisek


lbarcziova (approved):
thanks!

PR #436, opened 2023-01-19
Smaller repository_cache_storage

jpopelka:
4Gi for both prod and stg should be enough given that all worker pods say:
```
$ du -h /repository-cache/
20K     /repository-cache/
```

Until packit/packit-service#1840 is fixed.

### DIFF
--- a/vars/packit/prod_template.yml
+++ b/vars/packit/prod_template.yml
@@ -43,7 +43,7 @@ with_flower: true
 # with_pushgateway: true
 
 # with_repository_cache: true
-repository_cache_storage: 10Gi
+# repository_cache_storage: 4Gi
 
 with_fluentd_sidecar: true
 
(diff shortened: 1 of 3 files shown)


TomasTomecek (approved):

vars/packit/prod_template.yml
```diff
@@ -43,7 +43,7 @@ with_flower: true
 # with_pushgateway: true
 
 # with_repository_cache: true
-repository_cache_storage: 10Gi
+# repository_cache_storage: 4Gi
```
lachmanfrantisek, line 46: Have you wanted to comment this line?

lachmanfrantisek (commented):
:+1: 

Thanks! And sorry for not getting to that earlier.

vars/packit/prod_template.yml
jpopelka: Yes, previously the default was 16Gi, while the prod redefined it to 10Gi.
Now the default is 4Gi and it's used everywhere, including here in prod.
(Commented out line means "we use a default valued, defined somewhere else like this".)

PR #440, opened 2023-01-31
Fix flower-htpasswd secret notify -> nginx handler

jpopelka:


FrNecas (approved):
Thanks for fixing this

PR #441, opened 2023-02-01
Burstable QoS for postgres

jpopelka:
When I temporarily raised the mem limit to 2Gi on packit-prod a week ago the db became A LOT more responsive, (probably because the `shared_buffers` and `effective_cache_size` are set automatically based on the mem limit, see 2f07b23 commit message) while the total mem consumption stayed pretty much the same (maybe even a bit lower).

We can either play again (2f07b23) with setting the buffers & cache to higher values while keeping the pod in Guaranteed QoS (requests == limits). Or simply increase the limit permanently, which would change the QoS from Guaranteed to Burstable (limits > requests). I decided for the latter (see 3b0d20e for more info).

### DIFF
--- a/openshift/postgres.yml.j2
+++ b/openshift/postgres.yml.j2
@@ -52,13 +52,12 @@ spec:
 #            successThreshold: 1
 #            timeoutSeconds: 1
           resources:
-            # requests and limits have to be the same to have Guaranteed QoS
             requests:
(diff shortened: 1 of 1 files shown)


TomasTomecek (approved):
As I recall, that's indeed what Red Hat's psql container image does: it computes shared_buffers and effective_cache_size on the fly.

I don't personally have enough data to understand which of burstable or guaranteed is better, but I trust your judgement. Let metrics charts decide :)

PR #442, opened 2023-02-01
Import fedmsg image in fedora-source-git as well

jpopelka:
I've just realized the fedora-source-git-prod was trying to pull old fedmsg image, because the cron-job hadn't import a newer one into the image stream.


jpopelka:
Included in #443

PR #447, opened 2023-02-08
Workers: add liveness probe

jpopelka:
```
$ celery -A packit_service.worker.tasks status
->  celery@packit-worker-long-running-1: OK
->  celery@packit-worker-short-running-0: OK
2 nodes online.
```

The command usually takes about 7s to finish but can go up to 20-30s so the timeout is set to 1min to be on the safe side and not restart a worker needlessly.

There is no readiness probe because it IMHO makes no sense for pods with no inbound traffic.

We already had the probes a long time ago, but we removed them (94cee4a03) because they hadn't worked as expected.
Let's give it another try.

Related to packit/packit-service#1858

### DIFF
--- a/openshift/packit-worker.yml.j2
+++ b/openshift/packit-worker.yml.j2
@@ -123,6 +123,16 @@ spec:
             limits:
               memory: {{ worker_limits_memory }}
               cpu: {{ worker_limits_cpu }}
+          livenessProbe:
+            exec:
+              command:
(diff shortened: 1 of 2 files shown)


jpopelka:
recheck

lachmanfrantisek (approved):

TomasTomecek (approved):
👏🏼

PR #454, opened 2023-02-20
Double Postgresql memory requests & limits

jpopelka:
it's been quite hungry since packit/packit-service#1787

and tune CPU requests of some workloads (they should still be enough in (almost) all cases)


jpopelka:
recheck

majamassarini (approved):

PR #457, opened 2023-02-21
Increase liveness probe values

jpopelka:
We need to be much less aggressive because the probe is still failing quite often.

The cause might be the [`solo` pool, which blocks the worker while it executes tasks](https://docs.celeryq.dev/en/latest/userguide/workers.html#remote-control).
For now, just try a longer timeout (5min should be enough for all tasks) and if that doesn't help, we can go back to `prefork` pool.

### DIFF
--- a/openshift/packit-worker.yml.j2
+++ b/openshift/packit-worker.yml.j2
@@ -129,10 +129,11 @@ spec:
                 - bash
                 - '-c'
                 - 'celery -A $APP status | grep "$(uname -n): OK"'
-            # Start after 10m, every 5m with 1m timeout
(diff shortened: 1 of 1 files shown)


lachmanfrantisek (approved):

jpopelka:
There were still some failed probes overnight so I switched `solo` -> `prefork` and give it some time.

lbarcziova (approved):

jpopelka:
There has been no failed probe after I changed the pool to `prefork`, so looks like the `solo` pool had really been the cause.

Closing this in favor of packit/packit-service#1928 - please review there.

jpopelka:
OTOH, let's do both to be on the safe side.

PR #466, opened 2023-03-20
Allow prod postgres to utilize up to 1 cpu

jpopelka:
From the Openshift metrics, it looks like it needs around 0.2 cpu even at spikes but the queries that sort results improve a lot when the limit is set to 1 cpu.


TomasTomecek (approved):
let's try
do we wanna get this to prod tomorrow?

mfocko (approved):

PR #469, opened 2023-03-28
Label resources for MP+

jpopelka:
`shard: internal` - be available only on the corporate network
`shard: external` - available on the external/public network
`shard: ext-waf` - external/public network via Akamai (Web Application Firewall)


TomasTomecek (approved):
do we need some extra perms for the akamai shard? otherwise lgtm

PR #471, opened 2023-04-03
Create sandbox namespace the MP+ way

jpopelka:


openshift/sandbox-namespace.yml.j2
```diff
@@ -2,10 +2,22 @@
 # SPDX-License-Identifier: MIT
 
 ---
+{% if managed_platform %}
+apiVersion: tenant.paas.redhat.com/v1alpha1
+kind: TenantNamespace
```
TomasTomecek, line 7: interesting

TomasTomecek (approved):

PR #481, opened 2023-04-25
feat: switch deployment from RSA to ED25519

mfocko:


jpopelka (commented):
I see more places:

* https://github.com/packit/deployment/tree/main/secrets#what-secret-files-the-deployment-expects
* https://github.com/packit/deployment/blob/main/roles/generate-secrets/files/ssh_config#L4

There's also [playbooks/zuul-secrets.yml](https://github.com/packit/deployment/blob/main/playbooks/zuul-secrets.yml), but I don't think we actually use that. Let's not touch it and we'll see :)

> I have a feeling that this should wait for Monday, cause it might cause issues redeploying the prod, e.g. for the TLS certs

The [docs](https://github.com/packit/deployment/blob/main/docs/tls-certs.md#5-re-deploy-secrets-for-all-services-and-environments) suggests using `scripts/update_oc_secret.sh` which should not interact with this. Or I can refresh the certs now to unblock you :)

jpopelka:
TLS certs have been renewed

mfocko:
recheck

jpopelka (approved):

PR #483, opened 2023-05-02
Mpp migration

mfocko:


playbooks/import-images.yml
```diff
@@ -21,29 +21,29 @@
       changed_when: false
 
     - name: Check/import newer fluentd image into image stream
-      ansible.builtin.command: oc import-image is/fluentd:{{ deployment }} -n {{ service }}-{{ deployment }}
+      ansible.builtin.command: oc import-image is/fluentd:{{ deployment }} -n {{ service }}--{{ deployment }}
```
jpopelka, line 24: We use the `import-images` cron job (which runs this playbook) only on prod (on stage the importing is automatic) so you can skip this commit for now and we can do it once we're moving prod.
I'll update it meanwhile.

jpopelka:
Rebased, applied my comments, force pushed ;-)

But it still looks like history hasn't changed??? What am I missing?

Created #503 to finish it there.

PR #488, opened 2023-05-10
Set SQLALCHEMY_WARN_20 to indicate legacy patterns

jpopelka:
[SQLAlchemy 2.0 was released](https://www.sqlalchemy.org/blog/2023/01/26/sqlalchemy-2.0.0-released/) in January 2023 and one of the steps in the [Migration Guide](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-to-2-0-step-two-turn-on-removedin20warnings) is to set `SQLALCHEMY_WARN_20=1` env. var. to see deprecation warnings.

### DIFF
--- a/openshift/packit-service.yml.j2
+++ b/openshift/packit-service.yml.j2
@@ -66,6 +66,10 @@ spec:
             - name: SYSLOG_HOST
               # localhost doesn't work
               value: "127.0.0.1"
+{% if deployment != 'prod' %}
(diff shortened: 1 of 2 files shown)


jpopelka:
If we move to CentOS Stream 9 we'll stay with SQLAlchemy 1.4 for quite a long time, so we don't need this.
I can just set it in via the console and see if there are any issues.
Closing

PR #489, opened 2023-05-16
Fix ansible-lint naming-related errors

jpopelka:
- role-name: Role name generate-secrets does not match ``^[a-z][a-z0-9_]*$`` pattern.
- role-name[path]: Avoid using paths when importing roles.
- var-naming[no-role-prefix]: Variables names from within roles should use role_name_ as a prefix.
- var-naming[no-reserved]: Variables names must not be Ansible reserved names.


nforro (approved):
LGTM

PR #495, opened 2023-05-30
Fix keylime_server reference for internal_tf

majamassarini:


secrets/packit/prod/packit-service.yaml.j2
```diff
@@ -21,7 +21,7 @@ enabled_projects_for_internal_tf:
   - github.com/jontrossbach/xdp-tools
   - github.com/neoave/mrack
   - github.com/freeipa/freeipa
-  - github.com/linux-system-roles/keylime_server/
+  - github.com/linux-system-roles/keylime_server
```
mfocko, line 24: I guess we should probably handle it in the code :thinking: low-prio anyways :D

mfocko (approved):

PR #500, opened 2023-06-29
fix: adjust the docs for the packit.dev

mfocko:
Related to packit/packit.dev#679


lachmanfrantisek (approved):
Wow, thanks for going through that!

PR #515, opened 2023-07-28
fix(c9s): enable postgresql module

### LINKED ISSUE #514: [MP+] Daily backup of the database fails because of the version mismatch
Sentry Issue: [PCKT-002-PACKIT-SERVICE-63C](https://red-hat-it.sentry.io/issues/4327319283/?referrer=github_integration)

```
PackitCommandFailedError: Command 'pg_dump --file=/tmp/packit--stg_database_packit.sql --dbname=postgresql://packit-boss:‹censored›@postgres:5432/packit' failed.
  File "celery/app/trace.py", line 760, in __protected_call__
    return self.run(*args, **kwargs)
  File "packit_service/worker/tasks.py", line 511, in database_maintenance
    backup()
  File "packit_service/worker/database.py", line 126, in backup
    dump_to(file)
  File "packit_service/worker/database.py", line 108, in dump_to
    run_command(cmd=cmd)
  File "packit/utils/commands.py", line 128, in run_command
    raise PackitCommandFailedError(
```

Tried it myself:
* server: 15.2
* client: 13.11

mfocko:
With switch to MP+ we have also switched from PostgreSQL 13 to 15, this becomes an issue, since by default the C9S includes version 13 and regular backup is done from the worker.

Since we have multiple worker images, enable the ‹postgresql› module globally to reduce unnecessary copy-pasta into the playbooks for building workers.

Fixes #514

### DIFF
--- a/containers/Containerfile.c9s
+++ b/containers/Containerfile.c9s
@@ -8,6 +8,7 @@ ENV ANSIBLE_PYTHON_INTERPRETER=/usr/bin/python3 \
 
 RUN dnf install -y epel-release 'dnf-command(config-manager)' && \
     dnf config-manager --set-enabled crb && \
+    dnf module enable postgresql -y && \
     dnf update -y && \
     dnf install -y ansible python3-pip && \
     dnf clean all


jpopelka:
> we have multiple worker images

Which do you mean? p-s worker and hardly worker? staging and stable?

> Fixes #514

And will it work with server running v13 and pg_dump v15?
Because that'll be the situation on `packit-prod` once we built & import `packit-worker:prod` from this.

If not, then I think the easiest would be to revert `postgres` on `packit--stg` back to v13 and create a card for switching to v15 once everything is deployed to MP+.

mfocko:
> Which do you mean? p-s worker and hardly worker? staging and stable?

Packit Service and Hardly, also it feels rather easy to forget about it in 2 places.

> And will it work with server running v13 and pg_dump v15? Because that'll be the situation on `packit-prod` once we built & import `packit-worker:prod` from this.

I'll spin up a pod on prod to check it, though I have a feeling it's about backward-compatibility issues, so newer _server_ is a problem whereas the newer _dump_ shouldn't be.

> If not, then I think the easiest would be to revert `postgres` on `packit--stg` back to v13 and create a card for switching to v15 once everything is deployed to MP+. EDIT: Or move `postgres` on `packit-prod` to v15 together with this change.

I think it would be easier to just push the postgres to 15 on prod, rather than 15 to 13. Especially since postgres takes care of the “migration”.

mfocko:
Tested client/`pg_dump` of newer version than server, works fine.

jpopelka (approved):

mfocko:
regate

PR #522, opened 2023-08-21
Increase memory limit for packit-service to 4Gi

lbarcziova:
We agreed on increasing it permanently to avoid problems with database migrations.


jpopelka (approved):

lbarcziova:
regate

lbarcziova:
recheck

PR #523, opened 2023-08-21
Fix special chars in config & allow keylime-selinux

majamassarini:


secrets/packit/prod/packit-service.yaml.j2
```diff
@@ -31,7 +32,7 @@ command_handler_pvc_volume_specs:
   - path: /repository-cache
     pvc_from_env: SANDCASTLE_REPOSITORY_CACHE_VOLUME
     read_only: true
-# [TODO]: Switch to ‹aws-ebs› during migration of prod to MP+
+# [TODO]: Switch to <aws-ebs> during migration of prod to MP+
```
mfocko, line 35: Why is that an issue »now« and not before? :eyes:

mfocko (approved):

secrets/packit/prod/packit-service.yaml.j2
mfocko: Oh I know… the “OpenShift Dashboard”, we don't include the comments when it's run via `make deploy`.

majamassarini:
regate

majamassarini:
recheck

PR #528, opened 2023-09-01
fix: factor out deployment facts

mfocko:
To enable including them only in deployment, and therefore allow rendering of templates where we don't need to have all of the ansible facts related to deployment.


lachmanfrantisek (approved):
I believe you..;)

PR #542, opened 2023-11-14
feat: use staging TF API on stage

mfocko:
As it was suggested by @thrix (@mvadkert), we should use staging API of Testing Farm on our staging instance to test and validate their changes.

Given the fact there are not many users of Packit staging instance apart from Packit Team and Packit Validation, we should be able to verify the TF changes before they land on **both** TF and Packit production, which should in turn increase overall stability of Testing Farm experience while using Packit.


lbarcziova (approved):
thanks!

lachmanfrantisek (commented):
:+1:

PR #544, opened 2023-12-07
Allow github.com/ptoscano/insights-client to use internal TF

lachmanfrantisek:
Requested by @ptoscano


ptoscano:
> I assume this will be eventually replaced with `github.com/RedHatInsights/insights-client`. Or maybe we should add it right away?

That is correct; I can send a request for that separately (to cover also another project).

lachmanfrantisek:
> That is correct; I can send a request for that separately (to cover also another project).

I've added the other two as well.

ptoscano:
> > That is correct; I can send a request for that separately (to cover also another project).
> 
> I've added the other two as well.

Thank you!

PR #550, opened 2024-01-16
Specify authors in post output by move-stable

lbarcziova:


scripts/move_stable.py
```diff
@@ -359,6 +359,7 @@ def create_blogpost(
         "---\n"
         f"title: {title_text} in Packit\n"
         f"date: {today.strftime('%Y-%m-%d')}\n"
+        f"authors: <author>"
```
nforro, line 362: I wonder if you could get it dynamically from git config. If not, it would probably be nice to add a hint above not to forget to replace `<author>` with your name.

mfocko: If we all use RH mail, then yes, cause I set up the list of authors with Kerberos logins… but that's not foolproof :D though I think that test deployment should catch it :thinking: not sure now if it just dumps the name or complains

lbarcziova: I believe all of us use their RH mail :D PTAL

nforro (approved):

PR #553, opened 2024-01-26
chore: nuke the validation

mfocko:
Remove all validation related files, since they are no longer present after moving out the validation code to its own repository:

    https://github.com/packit/validation

All image builds and changes should be done there.


lbarcziova (approved):
thanks!

PR #558, opened 2024-03-07
fix: fix typos in the exceptions for limits

mfocko:
After switching our deployment to MP+ we (read as “me”) missed some occurences of hardcoded project/namespace for Packit Service deployment.

* Fix those typos.
* Add warning to the template.


majamassarini (approved):
Good catch! Thanks!

PR #559, opened 2024-03-29
Allow github.com/theforeman/foreman-installer to use internal TF

nforro:
Requested by @ekohl.

### DIFF
--- a/secrets/packit/prod/packit-service.yaml.j2
+++ b/secrets/packit/prod/packit-service.yaml.j2
@@ -33,6 +33,7 @@ enabled_projects_for_internal_tf:
   - github.com/ptoscano/insights-client
   - github.com/RedHatInsights/insights-client
   - github.com/RedHatInsights/rhc
+  - github.com/theforeman/foreman-installer
 
 command_handler: sandcastle
 command_handler_work_dir: /tmp/sandcastle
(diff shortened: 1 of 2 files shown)


nforro:
recheck

nforro:
recheck

majamassarini (approved):
🙏

lachmanfrantisek (approved):

nforro:
recheck

ekohl:
I tried to rerun https://github.com/theforeman/foreman-installer/pull/921 but it still doesn't allow RHEL composes. It's unclear to me how I actually get my tests to run.

As an aside: it is also possible to build on RHEL but run the tests upstream on CentOS Stream? For our specific tests we care less that it's really compatible with RHEL and having completely visible logs to all upstream users is important.

nforro:
> I tried to rerun [theforeman/foreman-installer#921](https://github.com/theforeman/foreman-installer/pull/921) but it still doesn't allow RHEL composes. It's unclear to me how I actually get my tests to run.

I've commented on the PR.

> As an aside: it is also possible to build on RHEL but run the tests upstream on CentOS Stream? For our specific tests we care less that it's really compatible with RHEL and having completely visible logs to all upstream users is important.

Yes, I suppose something like this should work (in this case you don't need `use_internal_tf` of course):

```yaml
  - job: tests
    trigger: pull_request
    targets:
      rhel-8-x86_64:
        distros: [centos-stream-8]
      rhel-9-x86_64:
        distros: [centos-stream-9]
```

PR #560, opened 2024-04-11
fix(imagestream): use quay.io images for nginx

mfocko:
Use nginx image from quay.io instead of ghcr.io, since it doesn't get “rate-limited”.


lbarcziova (approved):
thanks!

PR #567, opened 2024-04-18
Customize shm size in postgres pod

majamassarini:
By default shm size is 64MB, *dashboard usage* page uses around 100MB.

I can reproduce the error described in packit/packit-service#2385 locally using podman compose and I can fix it by setting `shm_size: 128Mi` in `docker-compose.yml`.

I tried deploying these openshift changes on my local openshift cluster but even though the new volume is created and mapped the size didn't change.

`df -h /dev/shm` is always 64MB

I would like to test this on stg, to be sure the problem is not related with my local openshift cluster.

But, as far as I can understand it, does exist a `SizeMemoryBackedVolumes` k8s *feature gate*, which is not enabled in the cluster by default (probably because this is a  dangerous feature https://github.com/kubernetes/kubernetes/issues/119611).
If this is not enabled in the cluster, resizing can not occur.

If this PR will not work on stg I think we have 2 solutions:
- ask to the cluster's maintainers if they have a way to increase shared memory for our pod
- redesign db queries in usage page

### DIFF
--- a/openshift/postgres.yml.j2
+++ b/openshift/postgres.yml.j2
@@ -58,12 +58,12 @@ spec:
 #                - /usr/libexec/check-container
           resources:
             requests:
-              memory: "{{ '2Gi' if project == 'packit--prod' else '256Mi' }}"
+              memory: "{{ '1Gi' if project == 'packit--prod' else '256Mi' }}"
               cpu: "30m"
             limits:
 # The shared_buffers & effective_cache_size values are set by the container
 # based on this limit, so keep the limit reasonably high.
-              memory: "{{ '4Gi' if project == 'packit--prod' else '512Mi' }}"
+              memory: "{{ '1536Mi' if project == 'packit--prod' else '512Mi' }}"
 # Not utilized most of the time, but useful during migrations and for some queries
               cpu: "1"
           volumeMounts:
--- a/openshift/postgres.yml.j2
+++ b/openshift/postgres.yml.j2
@@ -35,6 +35,11 @@ spec:
           ports:
             - containerPort: 5432
               protocol: TCP
+          # increase shm size, default is 64M
+          # dashboard usage page requires around 100MB
+          volumeMounts:
+            - mountPath: /dev/shm
+              name: cache-volume
           livenessProbe:
             exec:
               command:
@@ -53,12 +58,12 @@ spec:
 #                - /usr/libexec/check-container
           resources:
             requests:
-              memory: "{{ '2Gi' if project == 'packit--prod' else '128Mi' }}"
+              memory: "{{ '2Gi' if project == 'packit--prod' else '256Mi' }}"
               cpu: "30m"
             limits:
 # The shared_buffers & effective_cache_size values are set by the container
 # based on this limit, so keep the limit reasonably high.
-              memory: "{{ '4Gi' if project == 'packit--prod' else '256Mi' }}"
+              memory: "{{ '4Gi' if project == 'packit--prod' else '512Mi' }}"
 # Not utilized most of the time, but useful during migrations and for some queries
               cpu: "1"
           volumeMounts:
@@ -75,6 +80,11 @@ spec:
         - name: postgres-config
           configMap:
             name: postgres-config
+        - name: cache-volume
+          emptyDir:
+            medium: Memory
+            sizeLimit: 128Mi
+
   replicas: 1
   strategy:
     type: Recreate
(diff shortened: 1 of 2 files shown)


mfocko (approved):
> `df -h /dev/shm` is always 64MB

AFAIU `/dev/shm` is memory-mapped. Not sure if you are able to override it like that.

I would say that one of the options would be increasing the memory (though we don't have much left… and we're just running 2/2 workers; we have one more worker than usual for `short-running`, but it's still less than what I set up for the redis upgrade a month ago or so).

> * ask to the cluster's maintainers if they have a way to increase shared memory for our pod

Or just more memory in general…

> * redesign db queries in usage page

that would be probably ideal as it is the only API endpoint causing issues; IMO it would be probably for better to go with raw SQL queries instead of ORM which adds complexity because of the abstractions present…

---

I've added a commit with the currently deployed resource requests/limits.

I would probably recommend backup of the stage database and dumping the production one in its place, cause I have doubts about reproducibility of this issue on the “small-scale” contents of the stage's database.

( You could probably shortcut this by testing directly on prod… I'm neither endorsing it, nor disapproving it :D )

(( Also be wary of the fact that if the cluster can't provide the resources, it leaves everything dead; You can see the current status on `Workloads → Deployments → ‹specific deployment you're changing› → YAML (tab) → all the way at the bottom ‹status› section`; `oc status` shows a tl;dr, I see more details in the web UI :/ ← **In case you decide to go for the testing on prod this is kinda crucial, cause you can cause “nice” outage :smile:** ))

majamassarini:
> I would probably recommend backup of the stage database and dumping the production one in its place, cause I have doubts about reproducibility of this issue on the “small-scale” contents of the stage's database.

Luckily it is really easy to check. I just need to enter the postgres pod with `rsh` and run `df -h /dev/shm`.
If the size is 64MB then we still have the problem and the PR is useless and we should decide for one of the other two solutions.

majamassarini:
> I would say that one of the options would be increasing the memory (though we don't have much left… and we're just running 2/2 workers; we have one more worker than usual for `short-running`, but it's still less than what I set up for the redis upgrade a month ago or so).

The problem seems not to be related with memory in general. My local pods had plenty of memory and until the shared memory was 64MB the exception occurred. If we are not able to increase the shared memory we can not solve it, I fear.

> > * redesign db queries in usage page
> 
> that would be probably ideal as it is the only API endpoint causing issues; IMO it would be probably for better to go with raw SQL queries instead of ORM which adds complexity because of the abstractions present…

I remember having already worked on the raw queries once, at the time we decided to stay with the ORM. But if we have no other solutions, probably this will be the easiest way. And probably for the usage pages we can also use `views`...

majamassarini:
This PR was not working because the `deploy` target is not mounting the created volume. If you mount manually the volume then the shm is resized.

openshift/postgres.yml.j2
```diff
@@ -53,12 +58,12 @@ spec:
...
           resources:
             requests:
-              memory: "{{ '2Gi' if project == 'packit--prod' else '128Mi' }}"
+              memory: "{{ '1Gi' if project == 'packit--prod' else '256Mi' }}"
               cpu: "30m"
             limits:
 # The shared_buffers & effective_cache_size values are set by the container
 # based on this limit, so keep the limit reasonably high.
-              memory: "{{ '4Gi' if project == 'packit--prod' else '256Mi' }}"
+              memory: "{{ '1536Mi' if project == 'packit--prod' else '512Mi' }}"
```
jpopelka, line 66: I don't see anywhere explained why you've lowered the memory limit so I'm just wondering :-)

mfocko: Well…

1. we missed it during production migration to MP+, so in these conditions we had hard-coded `packit-prod` from Automotive :D
2. there were some issues with Celery (it somehow didn't register pods being killed off during rescaling, so there were new pods, but no tasks processed; same thing as with the first try to migrate prod to MP+ and even back to Automotive), so to mitigate the delays I scaled more pods (2 short-running + 2 long-running)
3. given more workers when I tried to redeploy prod with the correct resource requirements/limits I got hit with quotas :) fun fact: if you don't have enough resources, instead of falling back to the last deployment, OpenShift leaves everything that's affected by the changes dead :dead_inside:
4. and that's how I got to these funky numbers; now that I think about it, I should've run simplex algorithm on this :D

jpopelka: Ok, got it. I'll keep fingers crossed it's enough (I'd bet that your prod postgres would need more these days, but I've been probably optimizing since I left ;-) )

mfocko: well the fun thing is that the only thing “heating the place up” are usage pages, and that seems to be mitigated by the `/dev/shm` on separate volume

requesting more memory is an option, though the firewall is a bigger issue with the onboarding cards, so we're looking at moving to Fedora Infra… MP+ could be nice for CentOS/RHEL integration with stricter ACLs and the benefit of VPN access (because of stricter access control)

PR #568, opened 2024-04-22
fix(scripts): warn about the vars dependency

mfocko:
When rendering the secrets, e.g., config template, via script, we run some parts of the playbooks that may depend on the variables for a deployment (like MP+ vs Automotive, and such). Warn about this fact to make sure there are no discrepancies introduced by outdated “secrets”.


nforro (approved):
:pray:

lbarcziova (approved):

PR #571, opened 2024-05-08
Lower cpu requests

majamassarini:
The previous cpu limit increase 7b88605f16ee5a7f52ad06bef4d462eab940d645 should be enough.
The cluster running our pods should have enough resources to satisfy our needs without us blocking the resources.


mfocko (approved):
If you feel like we need more, you can bump the »limits_cpu«

jpopelka (approved):

PR #576, opened 2024-05-23
Expose Sentry related env var in dashboard pod

lbarcziova:
Related to packit/dashboard#410

### DIFF
--- a/openshift/dashboard.yml.j2
+++ b/openshift/dashboard.yml.j2
@@ -34,6 +34,8 @@ spec:
           env:
             - name: DEPLOYMENT
               value: {{ deployment }}
+            - name: VITE_SENTRY_DSN
+              valueFrom: {secretKeyRef: {name: sentry, key: dsn}}
           resources:
             requests:
               memory: "128Mi"


lbarcziova:
I see the PR actually requires 2 env vars, I will have a look into the SENTRY_AUTH_TOKEN.

mfocko (commented):
Why isn't DSN enough? :eyes:

Venefilyn:
> Why isn't DSN enough? 👀

@mfocko `SENTRY_AUTH_TOKEN` is to provide sourcemaps to Sentry so that it can decipher production builds, as they are minimized

mfocko:
Correct me if I'm wrong, but based on the docs:

- https://www.npmjs.com/package/@sentry/vite-plugin#authtoken
- https://docs.sentry.io/platforms/javascript/guides/gcp-functions/sourcemaps/uploading/vite/#configuration

I understand that this token needs to be present during the **build**, not **while** deployed, so this is something that should be exposed in the dashboard repo **and** for the GH Action (that actually builds the images).

Based on the docs from the first link, I have a feeling that it should work correctly by matching the commit SHA to the image.

(Based on the discussion with @lbarcziova)

Venefilyn:
> I understand that this token needs to be present during the build, not while deployed

Correct, should only be needed during build

lbarcziova:
thanks @Venefilyn for confirmation! Then I updated this PR only adding the VITE_SENTRY_DSN and the SENTRY_AUTH_TOKEN will be added in the dashboard repo.

mfocko (approved):

PR #589, opened 2024-08-01
fix: drop tokman deployment

mfocko:
* Drop deployment of tokman
* Mount GitHub App private key
* Adjust the variables in templates

TODO:

- [x] Updated `extra-vars.yml` in Bitwarden

Fixes packit/tokman#72


lbarcziova (approved):
thanks!

PR #590, opened 2024-08-06
feat(centpkg-sig): override the default config

mfocko:
Override the default config, since we are going to require the usage of GitLab.com for more streamlined experience across both CentOS Stream and CentOS SIGs.

Fixes packit/packit#2305


mfocko:
The fact I've added two identical files makes me want to touch the playbooks in a naughty way…

PR #598, opened 2024-09-23
fix(move-stable): remove archived repos

mfocko:
1. Remove tokman as it is no longer deployed.
2. Remove hardly as it is no longer deployed either.


lbarcziova (approved):
:pray:

Issue #593, opened 2024-08-21
Missing firewall rules, pt. 2


mfocko:
Issue for tracking the domains that have been discovered **not to** work on MP+.

### TODO

- [x] Request the firewall rules to be set
- [x] Verify they work
- [ ] Clone the issue for the next “waiting list”

### Waiting list

* https://imagemagick.org (by @lbarcziova for https://src.fedoraproject.org/rpms/ImageMagick/pull-request/13)
* libarchive.org (by @lbarcziova for https://dashboard.packit.dev/jobs/pull-from-upstream/10913)

### Notes

This is a tracking issue for any domains that are not allowed on the firewall and are blocking onboarding of Packit or Packit integration with an upstream forge. To not make the waiting requests “rot” in the backlog, there is a monthly deadline, in case there are no rules to be enabled, just shift the deadline by one more month, otherwise open a ticket to the IT to set the firewall rules.

### How to request the firewall rules

See [here](https://packit.dev/deployment/specifics/clusters#firewall)

mfocko:
Opened RITM1897456

mfocko:
Verfied. Works.

PR #599, opened 2024-09-25
docs(mpp): document MP+ specifics and firewall rules

mfocko:
Related to #593

### DIFF
--- a/docs/deployment/specifics/clusters.md
+++ b/docs/deployment/specifics/clusters.md
@@ -1,0 +1,62 @@
+---
+title: Clusters
+---
+
+## Automotive → MP+
+
+Currently we are moving from the Automotive cluster to the MP+ cluster. Our
+staging deployment is already present on the MP+ and there were some differences
+introduced once the migration started.
+
+### Namespaces
+
+Previously we have used a rather simple `packit-{{ deployment }}` namespace. On
+MP+ we are given `packit` as a tenant prefix, that needs to be separated from
+our custom namespaces by `--`, therefore we use `packit--stg` and
+`packit--stg-sandbox`.
+
+### Postgres
+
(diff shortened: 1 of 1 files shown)


docs/deployment/specifics/clusters.md
```diff
@@ -0,0 +1,62 @@
...
+Outgoing connections on the MP+ are implicitly denied and firewall rules must be
+explicitly requested. `10.0.0.0/8` network access from within our sandbox is
+explicitly denied.
+
+#### Requesting firewall rules
+
+1. Search for “egress firewall rules OSD” on Source
+2. Shiny link _opening a ticket_
+3. You can take an inspiration from RITM1861658; almost all of our tickets for
+   firewall rules follow a template below.
```
majamassarini, line 40:
```suggestion
   firewall rules follow the template below.
```

docs/deployment/specifics/clusters.md
```diff
@@ -0,0 +1,62 @@
...
+```
+• source: ‹list the clusters here›
+• destination: ‹list of domains›
+• protocol: 80, 443
+• justification: ‹almost as our release notes›
+(N clusters × M domains × P ports)
+```
+
+If you want to include subdomains, I would recommend typing it as
+`.example.com`, e.g., `.kde.org`.
```
majamassarini, line 53:
```suggestion
`.example.com`, e.g. `.kde.org`.
```

majamassarini (approved):

docs/deployment/specifics/clusters.md
mfocko: Oxford comma :P can be there, but doesn't have to :D

ok ok … I could've used it few lines below, but didn't, you win :D

PR #601, opened 2024-10-01
chore: use k8s_apply=true by default

### LINKED ISSUE #360: [Spike-ish] Use kubernetes.core.k8s apply=true by default?
The [apply parameter docs](https://docs.ansible.com/ansible/latest/collections/kubernetes/core/k8s_module.html#parameter-apply) says: 
"`apply` compares the desired resource definition with the previously supplied resource definition, ignoring properties that are automatically generated. `apply` works better with Services than `force=yes`."

I don't quite understand how it's actually different from the default behavior (what are the "properties that are automatically generated"?), but @csomh says it helps when there are properties being removed in the object.

Investigate:
- What it actually does?
- Why [some tasks already use it](https://github.com/packit/deployment/blob/4192911bebc8a26c365cf4f98a38f5526f3de8ef/playbooks/deploy.yml#L308) and others not?
- Would it make sense to [default](https://github.com/packit/deployment/blob/4192911bebc8a26c365cf4f98a38f5526f3de8ef/playbooks/deploy.yml#L52) to `apply=true`?

mfocko:
• Summary

Initially I wanted to do ‹k8s_apply=false› for postgres, and key-value
databases (such as Redis, Redict, or Valkey), because deploying on prod
with ‹k8s_apply=true› caused redeployment of the postgres which caused
a small outage (~5 minutes).

Right now when I tried to redeploy stage multiple times in a row, none of
the deployed services got redeployed, hence I come to the conclusion that
there were some changes on the production deployment that were applied
back then.

• Context from https://github.com/packit/deployment/issues/360

  · What it actually does?

    In the simple terms, it makes sure that the definition that is to be
    deployed matches the one that is already deployed. The difference has
    already manifested few times, e.g., when @majamassarini was adjusting
    the `/dev/shm` for the postgres deployment (https://github.com/mfocko/deployment/commit/bedef2026c84ea00bb329799cc9bef81687fe88d), the change did
    not get deployed.

  · Why some tasks already use it (e.g., Redis/Redict, Flower secret) and
    others not?

    not sure

  · Would it make sense to default to ‹apply=true›?

    Yes, but at the same time, applying meaningless changes to critical
    services, e.g., postgres or Redis/Redict/Valkey, can cause smaller
    outages.

Fixes https://github.com/packit/deployment/issues/360

### DIFF
--- a/playbooks/deploy.yml
+++ b/playbooks/deploy.yml
@@ -54,7 +54,7 @@
     check_vars_template_diff: true
     deployment_repo_url: https://github.com/packit/deployment.git
     # used by a few tasks below
-    k8s_apply: false
+    k8s_apply: true
     tokman:
       workers: 1
       resources:
@@ -184,8 +184,6 @@
         - postgres
 
     - name: Deploy redis
-      vars:
-        k8s_apply: true
       ansible.builtin.include_tasks: tasks/k8s.yml
       loop:
         - "{{ lookup('template', '{{ project_dir }}/openshift/redis.yml.j2') }}"
@@ -194,8 +192,6 @@
         - redis
 
     - name: Deploy redict
-      vars:
-        k8s_apply: true
       ansible.builtin.include_tasks: tasks/k8s.yml
       loop:
         - "{{ lookup('template', '{{ project_dir }}/openshift/redict.yml.j2') }}"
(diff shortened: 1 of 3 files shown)


lbarcziova (approved):
lgtm, we can revisit this if it starts causing bigger outages on redeployments

PR #606, opened 2024-10-18
fix(mpp): tag resources

mfocko:


lbarcziova (approved):

mfocko:
I don't like the copy-paste… it needs an overhaul… maybe on Friday…

PR #627, opened 2024-12-11
fix: replace deploy playbook with role

mfocko:
TODO:
- [x] ansible-lint warnings for role
  - will be handled in a follow up
- [x] check if it's possible to move the `roles/` out to the top-level to reduce the nesting
- [x] run tests on TF to make sure the changes are compatible

Supersedes #626


majamassarini (approved):
Thanks! 🤞🏻🤞🏻🤞🏻

lbarcziova (approved):

PR #634, opened 2025-01-15
Remove cpu limits

majamassarini:
We are hitting several cpu limits and weird packit-service behaviour with it.
Many resources on internet suggest not to use cpu limits
(but instead adjust cpu requests if needed),
like this one:
https://medium.com/inside-sumup/are-kubernetes-cpu-limits-bad-a04430bf54e1


majamassarini:
I close this one in favour of another one created from packit:remove-cpu-limits branch.

PR #636, opened 2025-01-15
Increase cpu limits for postgres

majamassarini:
If we don't set cpu limits they are being set by default (500m) but this value is too low for some of our pods.
Keep cpu limits.
Increase postgres cpu limit.

### DIFF
--- a/openshift/dashboard.yml.j2
+++ b/openshift/dashboard.yml.j2
@@ -48,6 +48,7 @@ spec:
               cpu: "5m"
             limits:
               memory: "256Mi"
+              cpu: "50m"
       volumes:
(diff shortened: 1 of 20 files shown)


lbarcziova (approved):
we could increase the CPU limit to 2 (the max allowed by cluster iiuc) also for other pods if needed, but from the metrics it looks like those are usually not problematic

majamassarini:
> we could increase the CPU limit to 2 (the max allowed by cluster iiuc) also for other pods if needed, but from the metrics it looks like those are usually not problematic

yes, I discussed the same with @mfocko and we agreed that we should be ok with this limits for now 🤞🏻.

PR #637, opened 2025-01-16
Increase cpu limits for long running workers

majamassarini:
I just realized I forgot the long running workers and looking at the metrics we sometime hit the limit also here...


lbarcziova (approved):
:+1:

PR #642, opened 2025-02-07
Increase short running worker memory

majamassarini:
After the last memory increase, it seems the workers are using more cpu and executing much more tasks. So memory seems to be a real bottleneck.

Screenshot From 2025-02-07 10-32-56

Screenshot From 2025-02-07 11-23-01


mfocko (approved):
:rocket:

majamassarini:
recheck

PR #645, opened 2025-02-20
Lower the timeout in redis-like config

lbarcziova:
This is a timeout after which idle connections are closed. I would try to lower it to 10 mins instead of 30 and see if it helps with downtime for the issue of idle workers (packit/packit-service#2697).


mfocko (approved):
:rocket:

PR #658, opened 2025-05-19
feat: handle tokens via Bitwarden Vault

mfocko:
Fixes packit/private#41

TODO:

- [ ] Update or write new documentation in `packit/packit.dev`.
- [x] Update the format of logins for the production in Bitwarden

Fixes packit/private#41


mfocko:
Redeployed the stage; it seems that we have some inconsistencies when templating the k8s specs, cause it also redeployed postgres, therefore postponing the production redeployment till evening or night-time, so it doesn't cause bigger disruptions as it redeploys the postgres or valkey

PR #662, opened 2025-07-04
fix(tasks): do not fetch secrets for dev deployment

### LINKED ISSUE #661: Ansible playbook tries to use Bitwarden in local/dev deployment
### What happened? What is the problem?

Running `make render-secrets-from-template` locally should not ask for bitwarden to finish.

### What did you expect to happen?

secrets are generated without requiring bitwarden vault in `DEPLOYMENT=dev`

### Workaround

- Unlock bitwarden with personal vault.

mfocko:
Fixes #661

@mynk8 could you please verify that it works ok?

### DIFF
--- a/tasks/set-facts.yml
+++ b/tasks/set-facts.yml
@@ -53,7 +53,6 @@
     bw_uri: "ansible://{{ service }}/{{ deployment }}"
 
 - name: Fetch git forges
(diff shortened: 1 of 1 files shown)


mynk8:
Yes, it works fine now, thanks !

PR #687, opened 2026-02-11
Adding conditional Log Detective variables to configuration templates

jpodivin:
I've used the same approach as utilized by other secrets. Both URL and token will be stored in bitwarden and rendered into template only if the URL is present. Log Detective interface server will always require authentication. Therefore, any configuration when URL is set, but the token is missing, is not viable.

Since removing the fields from configuration can be accomplished by just renaming, or removing, the variables in bitwarden, the fields were added to both prod and stage configuration templates.

RELEASE NOTES BEGIN

Log Detective URL and token are now conditional part of the service configuration. 

RELEASE NOTES END

### DIFF
--- a/secrets/packit/prod/packit-service.yaml.j2
+++ b/secrets/packit/prod/packit-service.yaml.j2
@@ -2375,3 +2375,8 @@ enabled_projects_for_fedora_ci:
   - https://src.fedoraproject.org/rpms/yt-dlp
   - https://src.fedoraproject.org/rpms/zathura-pdf-mupdf
   - https://src.fedoraproject.org/rpms/zeal
+
+# Log Detective interface server URL and token
+logdetective_enabled: false
+logdetective_url: {{ vault.logdetective.url }}
+logdetective_token: {{ vault.logdetective.token }}
--- a/secrets/packit/stg/packit-service.yaml.j2
+++ b/secrets/packit/stg/packit-service.yaml.j2
@@ -154,3 +154,8 @@ enabled_projects_for_fedora_ci:
   - https://src.fedoraproject.org/rpms/python-scikit-build-core
   - https://src.fedoraproject.org/rpms/python-specfile
   - https://src.fedoraproject.org/rpms/spglib
+
+# Log Detective interface server URL and token
+logdetective_enabled: false
+logdetective_url: {{ vault.logdetective.url }}
+logdetective_token: {{ vault.logdetective.token }}


secrets/packit/prod/packit-service.yaml.j2
```diff
@@ -2375,3 +2370,10 @@ enabled_projects_for_fedora_ci:
   - https://src.fedoraproject.org/rpms/yt-dlp
   - https://src.fedoraproject.org/rpms/zathura-pdf-mupdf
   - https://src.fedoraproject.org/rpms/zeal
+
+# Log Detective interface server URL and token
+{% if logdetective_enabled %}
+logdetective_enabled: {{ logdetective_enabled }}
+logdetective_url: {{ logdetective_url }}
+logdetective_token: {{ logdetective_token }}
+{% endif %}
```
nforro: Secrets stored in Bitwarden are available in the `vault` dict:

https://github.com/packit/deployment/blob/929e5c7a44c007b3675d9d318d03e5a91b77a8f3/roles/deploy/tasks/main.yml#L27-L32

It would also make more sense to use a hierarchical structure, for consistency with other services, for example:

https://github.com/packit/deployment/blob/929e5c7a44c007b3675d9d318d03e5a91b77a8f3/secrets/packit/prod/fedora.toml.j2#L47

secrets/packit/prod/packit-service.yaml.j2
```diff
@@ -2375,3 +2370,10 @@ enabled_projects_for_fedora_ci:
   - https://src.fedoraproject.org/rpms/yt-dlp
   - https://src.fedoraproject.org/rpms/zathura-pdf-mupdf
   - https://src.fedoraproject.org/rpms/zeal
+
+# Log Detective interface server URL and token
+{% if logdetective_enabled %}
```
nforro: Is `logdetective_enabled` supposed to be a secret as well?

nforro: > We already use `logdetective_enabled` in packit-service. https://github.com/packit/packit-service/blob/078f84ac6fd79dc8c47da815133871e3add5ef9d/packit_service/worker/handlers/koji.py#L365

How is that relevant here?

secrets/packit/stg/packit-service.yaml.j2
```diff
@@ -154,3 +154,8 @@ enabled_projects_for_fedora_ci:
   - https://src.fedoraproject.org/rpms/python-scikit-build-core
   - https://src.fedoraproject.org/rpms/python-specfile
   - https://src.fedoraproject.org/rpms/spglib
+
+# Log Detective interface server URL and token
+logdetective_enabled: {{ logdetective_enabled }}
```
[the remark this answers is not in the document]
jpodivin: Why not?

